PR #3, opened 2019-04-08
Add calling the closure from shared props using the Laravel container

rubenvanassche:
This will make it possible to add parameters to the shared props closure and let Laravel's container bind them.

For example:

```php
Inertia::share('user', function (Request $request) {
    return $request->user()->toArray();
});
```


reinink:
Dig it!

PR #5, opened 2019-04-09
Add Component methods as docblock properties to Intertia facade

patrickbrouwers:
Will be a great help for people using IDE's that want to use the Intertia facade and have autocomplete ont the methods.

I didn't provide any typehints on the method params, as the Component class also doesn't have them.


reinink:
Thanks! :)

Issue #7, opened 2019-04-18
Retrieve Inertia page's data in blade


rakk7:
We need a way to retrieve Inertia's page's data in blade. By adding a method like:

```php
Inertia::get('')
```

**For the readers:**
There is a lot of use cases for this, one of them would be when sharing a page in social networks.

For example, when sharing a page to Facebook, Facebook will go to your page and look for "open graph" metadata. In which Facebook's bots will not see ( since Facebook 's bots doesn't render JS ). Therefore, "open graph" metadata must be rendered server side for the first page load.

This is a known issue been around for SPA frameworks. And since Inertia is ( server-driven single page apps ), It solve this issue easily without any complicated workarounds.

https://github.com/nuxt/vue-meta/issues/133

--
For the ones who interested in SEO:
This also can be helpful to set the page's ( title - meta description - etc) in Blade ( for the first page load ). And then when a user do Inertia.visit, we can change the page's title like this:
```
let title = document.title;
if (title === this.page.props.meta.title) return;

title = this.page.props.meta.title;
```

rakk7:
made a PR https://github.com/inertiajs/inertia-laravel/pull/8

PR #8, opened 2019-04-18
add a method to retrieve page's data

rakk7:
closes: https://github.com/inertiajs/inertia-laravel/issues/7

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -26,6 +26,11 @@ class Component
         return Arr::set($this->sharedProps, $key, $value);
     }
 
+    public function get($key)
+    {
+        return Arr::get($this->sharedProps, $key);
+    }
+
     public function version($version)
     {
         $this->version = $version;


reinink:
So I like the idea here, and the use-cases you outline in #7 are good ones. That said, I'm not sure this implementation makes sense.

Does it really make sense to send data that may only be used in the server-side rendering with every Inertia request? This just bloats the JSON response unnecessarily.

It makes me wonder if we should consider something like this:

```php
class EventsController extends Controller
{
    public function show(Event $event)
    {
        return Inertia::render('Event', [
            // Only for JavaScript rendering
            'event' => $event->only('id', 'title', 'start_date', 'description'),
        ], [
            // Only for sever side rendering
            'meta_description' => $event->meta_description,
            'twitter_card' => 'summary_large_image',
            'twitter_image' => $event->twitter_image,
        ]);
    }
}
```

Admittedly, quite often there could be overlap between the two, for example, the title. Need to think on this a little.

rakk7:
I agree with you, I think we can add an optional method ```->with($data)``` for the data that will be only used for server-side. ( this can be a in separate issue topic )

```
return View::make($this->rootView, [
    'page' => [
        'component' => $component,
        'props' => array_merge($this->sharedProps, $props),
    ],
])->with($this->viewData);
```
---
On the other hand, we still need a way to access the same page's data in blade, such as the "title" case. In which this PR trying to do.

reinink:
Wait a minute, this already works:

```php
return Inertia::render('Event', ['event' => $event])
                ->with(['meta_description' => 'something']);
```

Which means you can get access to this data in your root template:

```blade
{{ $meta_description }}
```

That's because `Inertia::render()` returns a `View` instance. Which means, you can also get access to the props data this way:

```blade
{{ $page['props']['event']->title }}
```

We could still add `Inertia::get()` to make accessing the prop data easier, but really, this is more of a documentation issue.

I'm going to test this PR and see how it feels.

rakk7:
Just wow, how did i missed that! 💯IQ 😞

reinink:
Hah, it wasn't immediately obvious to me either. I'm going to add this to the docs.

reinink:
Merged, and added to docs: 0f7bbcc712eb58f0ddc8f8d64f9474f79809cdec

PR #10, opened 2019-04-25
use fullUrl() instead of url() method when assets change

cmizzi:
By now, the redirection when assets are changed also returns the query string.


reinink:
Good catch!

PR #14, opened 2019-04-26
inject props in the blade template

cmizzi:
The documentation is wrong by default. It tells that the `Inertia::get()` method can retrieve props, but only can retrieve shared props. This PR is injecting props into the blade template, to use them. It seems logic to keep the `view->with()` documentation, but `Inertia::get()` seems useless for the template part...

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -125,10 +125,10 @@ Inertia::share('auth.user', function () {
 
 ## Accessing data in root template
 
-There are situations where you may want to access your prop data in your root Blade template. For example, you may want to add a meta description tag, Twitter card meta tags, or Facebook Open Graph meta tags. To do this, use the `Inertia::get()` helper.
+There are situations where you may want to access your prop data in your root Blade template. For example, you may want to add a meta description tag, Twitter card meta tags, or Facebook Open Graph meta tags. To do this, all props are injected in the blade file, as regular Blade variable.
 
 ~~~blade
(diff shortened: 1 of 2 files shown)


reinink:
This is what happens when I merge in code that I haven't fully tested. 😂

Great find on this. I don't think the solution is 100% right though. You're merging the shared props, but this is actually already done here:

https://github.com/inertiajs/inertia-laravel/blob/db9a7b3f50e22103ec9bae9010dc656ad5272b1a/src/Component.php#L54

Meaning everything is technically available in the $page variable:

```blade
<meta name="twitter:title" content="{{ $page['props']['event']->title }}">
```

I say we just remove `Inertia::get()` entirely, and add this approach the docs.

Do you want to update this PR to reflect that?

cmizzi:
I didn't notice that. I'm on it !

cmizzi:
So ! Now, I removed `Inertia::get()` method. But, it could be interesting to retrieve shared data directly from the Inertia instance (like service, etc.), so I just move the logic into the `share($key, $value = null)` method.

cmizzi:
I have made the change, but commit doesn't appear (because of merged status ?). Should I create another one ?

reinink:
Hey @cmizzi! I merged in your stuff yesterday and then made the change myself. It's done. 👍

cmizzi:
Cool, thanks ! :+1:

PR #18, opened 2019-05-08
Use compact method

MichaelDeBoey:
Use `compact` method instead of using
```php
['page' => $page]
```

### DIFF
--- a/src/Component.php
+++ b/src/Component.php
@@ -67,6 +67,6 @@ class Component
             ]);
         }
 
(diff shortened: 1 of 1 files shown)


reinink:
Thanks for the idea @MichaelDeBoey, but I like the simplicity of the current implementation. 😛

MichaelDeBoey:
No prob, just wanted to suggest this since I see it used a lot in the Laravel community

PR #24, opened 2019-05-20
Add responsable response

### LINKED ISSUE #20: Accessing data in root template + inertia_vue issue!!
# laravel router 

```php 
Route::get('/contact', function () {
  Khead::setTitle('acontact');
  $event = "ok this is event";
  return Inertia::render('Contact', ['eventi' => $event])
                ->with(['meta_description' => $event]);
});
```
# inertia vue 

```vue 
<inertia-link href="/contact">contact</inertia-link>
```

`when I click contact route that time return an error` 

```bash 
BadMethodCallException
Method Illuminate\Http\JsonResponse::with does not exist.
```

`but normal anchor tag works properly.`

reinink:
This adds a new responsable Inertia response, which can have additional methods added to it (ie. `with()`.

For example, if you wanted to send Blade specific values:

```php
return Inertia::render('Dashboard/Index', $props)->with('foo', 'bar');
```

Not convinced this method should be called `with()`. I'm inclined to use that as an alternative way to set prop data, similar to how normal views work in Laravel. Maybe it should just be `withViewData()`. 🤔 

Resolves #20

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -24,7 +24,7 @@ class Response implements Responsable
         $this->version = $version;
     }
 
-    public function with($key, $value)
+    public function withViewData($key, $value)
     {
         $this->viewData[$key] = $value;
 
--- a/src/Response.php
+++ b/src/Response.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Inertia;
+
+use Illuminate\Http\JsonResponse;
+use Illuminate\Support\Facades\View;
+use Illuminate\Contracts\Support\Responsable;
+
+class Response implements Responsable
+{
+    protected $rootView;
+    protected $sharedProps;
+    protected $version;
+    protected $component;
+    protected $props;
+    protected $viewData = [];
+
+    public function __construct($rootView, $sharedProps = [], $version, $component, $props)
+    {
+        $this->rootView = $rootView;
+        $this->sharedProps = $sharedProps;
+        $this->version = $version;
+        $this->component = $component;
+        $this->props = $props;
+    }
+
+    public function with($key, $value)
+    {
+        $this->viewData[$key] = $value;
+
+        return $this;
+    }
+
+    public function toResponse($request)
+    {
+        $page = [
+            'component' => $this->component,
+            'props' => array_merge($this->sharedProps, $this->props),
+            'url' => $request->getRequestUri(),
+            'version' => $this->version,
+        ];
+
+        if ($request->header('X-Inertia')) {
+            return new JsonResponse($page, 200, [
+                'Vary' => 'Accept',
+                'X-Inertia' => 'true',
+            ]);
+        }
+
+        return View::make($this->rootView, $this->viewData + ['page' => $page]);
+    }
+}
--- a/src/Response.php
+++ b/src/Response.php
@@ -8,20 +8,20 @@ use Illuminate\Contracts\Support\Responsable;
 
 class Response implements Responsable
 {
+    protected $component;
+    protected $props;
     protected $rootView;
     protected $sharedProps;
     protected $version;
-    protected $component;
-    protected $props;
     protected $viewData = [];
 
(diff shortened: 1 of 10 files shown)


sebastiandedeyne:
I strongly agree that `Inertia::render` should behave as much like Blade views as possible, and in turn be an alternative way to set prop data.

Would it be too naive to just have one set of data that get shared with Inertia _and_ the Blade view?

Otherwise, `withViewData` sounds good 👍

reinink:
> I strongly agree that `Inertia::render` should behave as much like Blade views as possible, and in turn be an alternative way to set prop data.

I agree. So, then it should  be `withViewData()`. I think that's more clear anyway of what's actually happening.

> Would it be too naive to just have one set of data that get shared with Inertia and the Blade view?

I think so, only because you might have data that's intended just for Blade (ie. meta tag data), and passing that via XHR each time just means sending useless extra data over the wire, when 99.% of the time it won't be needed (except for on the initial page load).

src/Response.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Inertia;
+
+use Illuminate\Http\JsonResponse;
```
MichaelDeBoey: This adds the `illuminate/http`  dependency

src/ResponseFactory.php
```diff
@@ -53,20 +48,6 @@ public function render($component, $props = [])
...
-
-        if (Request::header('X-Inertia')) {
-            return Response::json($page, 200, [
-                'Vary' => 'Accept',
-                'X-Inertia' => 'true',
-            ]);
-        }
-
-        return View::make($this->rootView, ['page' => $page]);
+        return new Response($this->rootView, $this->sharedProps, $this->getVersion(), $component, $props);
```
rakk7: Uncle bob would say: too many args to be passed

src/Response.php
MichaelDeBoey: Right now we only have 
https://github.com/inertiajs/inertia-laravel/blob/e69451817be2700c66657f0e07e41be412716a40/composer.json#L25-L26

reinink: Hmm, but when use facades (which I do in this library), such as `Illuminate\Support\Facades\App` or `Illuminate\Support\Facades\View` or others, it technically requires additional dependencies regardless. Which has me wondering what the usefulness of listing these dependencies really is.

MichaelDeBoey: Don't know how other packages handle this kind of things tbh...

sebastiandedeyne: At Spatie we require every package we explicitly import, so in this case we'd require `illuminate/http` as a dependency.

There's also the nuclear option of just requiring `laravel/framework`, but that brings along other issues (I forgot which ones, but I remember regretting it at some point)

reinink: It just makes *no sense* to me to require `illuminate/support`, and then use the facades available in that package, but then omit the underlining dependencies for those facades. Things would blow up. I still don't get the point.

sebastiandedeyne: Because that's not our problem. Our code is only `illuminate/support`'s API. If `illuminate/support` is calling other things under the hood, it's their responsibility to require those packages.

reinink: Right, and that makes sense. I just mean _practically_ speaking, what advantage is any of this? How does it actually help anything?

The only reason I've even been inclined to include these kind of dependencies in the past is for testing purposes, because in the actual app...the Laravel dependencies are there. And to be honest, for testing, I've used `orchestra/testbench`, which does require `laravel/framework`.

sebastiandedeyne: Yeah your second paragraph is true. However, if you're testing the package without testbench, it makes sense to correctly declare the dependencies.

src/ResponseFactory.php
reinink: I changed the order of the params, putting the optional ones last. I think that's totally fine for a constructor.

PR #29, opened 2019-05-24
Add helper function inertia()

georgehanson:
When using the Laravel framework I personally find it cleaner to use the `view()`, so I'd also really like to be able to use the same approach with Intertia.

This pull request simply adds a helper function `inertia()`.

For example:

```php
return Intertia::render('Welcome', ['name' => 'George']);
```

becomes

```php
return inertia('Welcome', ['name' => 'George']);
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,10 @@
     "autoload": {
         "psr-4": {
             "Inertia\\": "src"
-        }
+        },
+        "files": [
+            "helpers.php"
+        ]
     },
     "autoload-dev": {
         "psr-4": {
--- a/helpers.php
+++ b/helpers.php
@@ -1,8 +1,12 @@
 <?php
 
 if (! function_exists('inertia')) {
-    function inertia($component, $props)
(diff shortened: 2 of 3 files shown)


georgehanson:
I have made a further change so that if you don't pass any parameters to the `inertia()` function, it returns an instance of the ResponseFactory, similar to how `view()` works in Laravel.

This allows you to do things such as:

```php
intertia()->share();
```

georgehanson:
@reinink any chance of reviewing this? Thanks :)

reinink:
@georgehanson Sorry, I've been super busy and behind on my open source stuff. Curious, why did you close this?

georgehanson:
@reinink no problem :)

I closed it be because of two main reasons. Firstly I want to just do it in one commit and secondly because when using it locally I noticed a couple of issues with my implementation. So I want to just iron out a few things. I'll probably be creating a new PR tomorrow

georgehanson:
PR resubmitted as #36.

PR #33, opened 2019-06-08
Move tests under main inertia namespace

sylarbg:
The current structure of the test namespace have problems with autocompletion, member resolution and inheritance analysis for IDEs like phpstorm. Тhe reason is multiple declarations of Tests\TestCase namepsace in inertia-laravel package and Tests\TestCase in standard Laravel app.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     },
     "autoload-dev": {
         "psr-4": {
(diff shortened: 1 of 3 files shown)


reinink:
Cool, thanks @sylarbg!

PR #36, opened 2019-06-09
Add inertia() helper function

georgehanson:
@reinink This is a re-open on my previous PR #29.

---------------------------------------------------

When using the Laravel framework I personally find it cleaner to use the `view()`, so I'd also really like to be able to use the same approach with Intertia.

This pull request simply adds a helper function `inertia()`.

For example:

```php
return Intertia::render('Welcome', ['name' => 'George']);
```

becomes

```php
return inertia('Welcome', ['name' => 'George']);
```

You can also access the `ResponseFactory` instance by not passing any parameters to the function. This allows you to make calls to other methods such as `share`.

```php
inertia()->share('name', 'George');
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,10 @@
     "autoload": {
         "psr-4": {
             "Inertia\\": "src"
-        }
+        },
+        "files": [
+            "./helpers.php"
+        ]
     },
     "autoload-dev": {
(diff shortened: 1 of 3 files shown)


reinink:
Dig it, thanks @georgehanson.

PR #39, opened 2019-06-14
Allow passing Arrayable objects

bonzai:
Instead of manually passing an array with data:

```php
class EventsController extends Controller
{
    public function show(Event $event)
    {
        return Inertia::render('Event', [
            'event' => $event->only('id', 'title', 'start_date', 'description'),
        ]);
    }
}
```

we can use an `Arrayable` object (for example, like `Collection`) to pass the data:

```php

class EventProps implements Arrayable
{
    public $event;

    public function __construct(Event $event)
    {
        $this->event = $event;
    }

    public function toArray()
    {
        return ['event' => $this->event->only('id', 'title', 'start_date', 'description')];
    }
}

class EventsController extends Controller
{
    public function show(Event $event)
    {
        return Inertia::render('Event', new EventProps($event));
    }
}
```

This way we can reuse the logic and simplify the code.

Similar functionality is already present in Laravel:

- https://github.com/laravel/framework/blob/5.8/src/Illuminate/View/View.php#L76
- https://github.com/laravel/framework/blob/5.8/src/Illuminate/View/Factory.php#L238-L241

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -6,6 +6,7 @@ use Closure;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\View;
+use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Support\Responsable;
 
 class Response implements Responsable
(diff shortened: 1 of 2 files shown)


reinink:
Thanks for the contribution!

PR #44, opened 2019-07-23
Use Response factory to build view

rodrigopedra:
Laravel `Responsable` trait has a type hint in its `toResponse` which states the return of this method should be a `\Symfony\Component\HttpFoundation\Response` object.

My use case is that I am using Inertia witth Spatie [Laravel View Models](https://github.com/spatie/laravel-view-models) package.

if I try to extend the view model `toResponse` method and return Inertia::render, I get an error because their package uses a return type hint.

Changing to use the Response facade instead of the view facade already wraps the view in a Request object and it aligns with the expected return type on the `Responsable` trait.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -3,11 +3,12 @@
 namespace Inertia;
 
 use Closure;
+use Illuminate\Support\Arr;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Support\Facades\App;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Support\Responsable;
-use Illuminate\Support\Facades\Response as ResponseFacade;
+use Illuminate\Support\Facades\Response as ResponseFactory;
 
 class Response implements Responsable
 {
@@ -52,7 +53,7 @@ class Response implements Responsable
         $only = array_filter(explode(',', $request->header('X-Inertia-Partial-Data')));
 
         $props = ($only && $request->header('X-Inertia-Partial-Component') === $this->component)
-            ? array_only($this->props, $only)
+            ? Arr::only($this->props, $only)
             : $this->props;
 
         array_walk_recursive($props, function (&$prop) {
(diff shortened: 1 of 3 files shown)


rodrigopedra:
Hi @reinink 

I merged the recent changes in the master branch.

If you can take a look in this PR I would really appreciate. Thanks!

rodrigopedra:
Hi @reinink and @sebastiandedeyne 

I merged the recent changes from the master branch into this patch.

I appreciate if you could take a look.

This time I also tried to keep the tests the most close to the original tests.

src/Response.php
```diff
@@ -74,6 +74,6 @@ public function toResponse($request)
             ]);
         }
 
-        return View::make($this->rootView, $this->viewData + ['page' => $page]);
+        return ResponseFacade::view($this->rootView, $this->viewData + ['page' => $page]);
```
sebastiandedeyne: Having to do this makes me think we should consider renaming `Inertia\Response` to `Inertia\InertiaResponse`.

What do you think @reinink?

MichaelDeBoey: Didn't we rename it before from `InertiaResponse` to `Response` so it wouldn't  be verbose anymore? 🤔

reinink: Yeah I'm not really a fan. Isn't that the whole point of namespaces? I get it though, it feels weird in this situation to have a `ResponseFacade`. 

I think the better naming here is rename `ResponseFacade` to `ResponseFactory`, because that's actually what is is.

```php
use Illuminate\Support\Facades\Response as ResponseFactory
```
```php
```

Thoughts?

rodrigopedra: Wouldn't it get confused with Inertia's own `Inertia\ResposeFactory`?

rodrigopedra: Renamed the Façade alias to `ResponseFactory` as suggested by @reinink

sebastiandedeyne: Sounds good 👍

rodrigopedra:
In the last commit I changed the `array_only` usage to the `Arr::only` from the `Illminate\Support\Arr` class. 

As the array and string helpers are no longer shipped with the core framework from the 6.0 release I thought it was a good idea to make this simple change.

reinink:
@rodrigopedra This actually just got changed! 😂

See #68

rodrigopedra:
@reinink , no problem

Just merged master again

reinink:
Thanks @rodrigopedra!

PR #46, opened 2019-07-29
Fix request session check

sebastiandedeyne:
`$request->session()` throws an exception if no session is set, so this check doesn't work at the moment.

`hasSession` returns a boolean instead.


reinink:
Good catch 👌

PR #51, opened 2019-08-12
Assertion Tests & Implementation

Juhlinus:
# Assertions in the Inertia Laravel Adapter
I'm pretty sure that most of us that are testing are making use of the neat macros that was made for PingCRM, so I thought I'd make a PR merging those assertions into the adapter.

# Motivations
I don't really want my TestCase.php to be bloated with a ton of macros, so I decided to make these assertions and override the default Laravel TestResponse with these. (You will of course be able to use all assertions that comes with Laravel as well.)

# Usage
Simply extend the Inertia TestCase instead of the default Laravel one.

## Before
```php
<?php

namespace Tests;

use Illuminate\Support\Arr;
use PHPUnit\Framework\Assert;
use Illuminate\Foundation\Testing\TestResponse;
use Illuminate\Foundation\Testing\TestCase as BaseTestCase;

abstract class TestCase extends BaseTestCase
{
    use CreatesApplication;

    protected function setUp(): void
    {
        parent::setUp();

        TestResponse::macro('props', function ($key = null) {
            $props = json_decode(json_encode($this->original->getData()['page']['props']), JSON_OBJECT_AS_ARRAY);
            if ($key) {
                return Arr::get($props, $key);
            }
            return $props;
        });

        TestResponse::macro('assertHasProp', function ($key) {
            Assert::assertTrue(Arr::has($this->props(), $key));
            return $this;
        });

        TestResponse::macro('assertPropValue', function ($key, $value) {
            $this->assertHasProp($key);
            if (is_callable($value)) {
                $value($this->props($key));
            } else {
                Assert::assertEquals($this->props($key), $value);
            }
            return $this;
        });

        TestResponse::macro('assertPropCount', function ($key, $count) {
            $this->assertHasProp($key);
            Assert::assertCount($count, $this->props($key));
            return $this;
        });
    }
}
```

## After

```php
<?php

namespace Tests;

use Tests\CreatesApplication;
use Inertia\Testing\TestCase as BaseTestCase;

abstract class TestCase extends BaseTestCase
{
    use CreatesApplication;
}
```

See? Much cleaner than before! 

### Docs

If this get merged I'll sit down and write some documentation for this as well. 😄

### DIFF
--- a/src/Testing/TestCase.php
+++ b/src/Testing/TestCase.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Inertia\Testing;
+
+use Inertia\Testing\TestResponse;
+use Illuminate\Contracts\Console\Kernel;
+use Illuminate\Foundation\Testing\TestCase as BaseTestCase;
+
+abstract class TestCase extends BaseTestCase
+{
+	/**
+	 * Create the test response instance from the given response.
+	 *
+	 * @param  \Illuminate\Http\Response $response
+	 *
+	 * @return \Illuminate\Foundation\Testing\TestResponse
+	 */
+	protected function createTestResponse($response)
+	{
+	    return TestResponse::fromBaseResponse($response);
+	}
+}
--- a/src/Testing/TestResponse.php
+++ b/src/Testing/TestResponse.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Inertia\Testing;
+
+use Illuminate\Support\Arr;
+use PHPUnit\Framework\Assert as PHPUnit;
+use Illuminate\Foundation\Testing\TestResponse as BaseResponse;
+
+class TestResponse extends BaseResponse
+{
+	public function assertHasProp($prop)
+	{
+		PHPUnit::assertTrue(Arr::has($this->props(), $prop));
+
+		return $this;
+	}
+
+	public function assertPropValue($prop, $value)
+	{
+		$this->assertHasProp($prop);
+
+		if (is_callable($value)) {
+			$value($this->props($prop));
+		} else {
+			PHPUnit::assertEquals($this->props($prop), $value);
+		}
+
+		return $this;
+	}
+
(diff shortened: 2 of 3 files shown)


reinink:
Thanks for this @Juhlinus, but I actually have some updated testing helpers. I'm going to leave this PR open and will post those helpers when I have a minute.

Juhlinus:
> Thanks for this @Juhlinus, but I actually have some updated testing helpers. I'm going to leave this PR open and will post those helpers when I have a minute.

Great! Looking forward to it!

herpaderpaldent:
i am curious about these new helpers too @reinink

Juhlinus:
> i am curious about these new helpers too @reinink
@herpaderpaldent Jonathan has a day-job and can't push out new changes everyday, unfortunately.

He will post an update in this thread when the new helpers are available. :)

m1guelpf:
This would probably make better sense as a trait instead of a class?

robsontenorio:
it would be nice!

nasyrov:
IMHO better to create a proper macros class and boot in the service provider

robsontenorio:
Macros would be a option. But auto completion would be a plus!

johanvanhelden:
But what if the user already has his own `BaseTestCase` that he uses with certain customizations. 
This way it is going to be extend-ception. 

Maybe a trait or macro would be a better option.

mojtabaahn:
Hi there. any updates on this ?

Juhlinus:
@mojtabaahn Not yet!

If you want to use it, then just add it to your project. 😄

claudiodekker:
Thanks @Juhlinus! We just merged #220, meaning this PR is no longer necessary 👍

Juhlinus:
> Thanks @Juhlinus! We just merged #220, meaning this PR is no longer necessary, so I'll be closing this one now 👍

Nice job buddy!

PR #53, opened 2019-08-15
remove semicolon inside inertia share array lazily example

markvaneijk:
Small syntax error in the example.


reinink:
Thanks Mark! 💪

PR #60, opened 2019-08-26
Adds a Route Macro

peterfox:
Adds the ability to do simple inertia responses with a single route line without needing to use a closure which wouldn't be supported by Laravel's route caching mechanism.

```
Route::inertia('Home')
Route::inertia('About', ['email' => 'test@test.com']);
```

So that there isn't much type checking, partly because I think it could be made more advance later on by potentially passing on classes to resolve from the container to handle the props.


Juhlinus:
Hello @peterfox !

This is a neat idea, but I don't personally see this is as something that should be added the adapter.

The reasoning behind this is that this seems like a quality-of-life feature as opposed to something that contributes something to the adapter itself.

The Laravel Framework itself is missing this feature as well, but have yet to implement a helper macro for it, which I believe shouldn't be done there either.

I would suggest maybe making a separate package that developers could pull into their project if they wish to use this feature.

Also, this doesn't seem to work out of the box.

I get the folllowing error when I try to use the `inertia` macro.
> Invalid route action: [App\Http\Controllers\Inertia\Controller].

`routes/web.php`
```php
Route::inertia('/home/', 'Home');
```
Neither of your examples match the signature of the macro, which is misleading. 

According to both your examples the first argument is the component, which doesn't match the signature, as well as giving the impression that the second argument can be left out:
```php
Router::macro('inertia', function ($uri, $component, $props = []) {
```

sebastiandedeyne:
I wouldn't mind this.

Laravel also has a `Route::view` method, this would be Inertia's counterpart.

https://laravel.com/docs/5.8/routing#view-routes

pedroborges:
I'm a big fan of `Route:view` and find it handy in some situations.

Juhlinus:
> I wouldn't mind this.
> 
> Laravel also has a `Route::view` method, this would be Inertia's counterpart.
> 
> https://laravel.com/docs/5.8/routing#view-routes

This totally slipped my mind, haha.

sebastiandedeyne:
Closure routes can't be cached 😬

peterfox:
@Juhlinus I've updated my PR which should fix the controller not being recognised (Laravel can do some weird namespacing by default which my tests didn't take into account). I've updated the confusing example in original comment as well.

Thank you for considering the idea. It was something I added to my own project and thought it might be useful as a lot of the time you might just be loading a simple page but having to make extra controllers for it.

PR #63, opened 2019-08-27
Adds the ability to switch off some components from the provider

peterfox:
I thought it might be useful as sometimes the end user will want to switch off particular parts of the package if they want to extend it further. By default everything is enabled but if the user wants to they can publish the config and disable the components.

I prefer this personally over having to disable auto discovery and extend the service provider for a package like this, but it's a valid alternative.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -1,0 +1,7 @@
+<?php
+
+return [
+    'use_middleware' => true,
+    'use_router_macro' => true,
+    'use_blade_directive' => true,
+];
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -2,18 +2,28 @@
 
 namespace Inertia;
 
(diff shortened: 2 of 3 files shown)


sebastiandedeyne:
While I get where this is coming from, it introduces a lot of code.

I'd rather wait until someone has a real problem, and see how we should continue then.

reinink:
Yup, exactly my thoughts as well @sebastiandedeyne.

Thanks Peter! I really do appreciate the interest, and time spent, but I don't really think this is is needed. There really is no harm in having the router macro and Blade directive. The only questionable one is maybe the middleware, but even there, since it only runs if the `X-Inertia` header is present, I really don't think it will ever be an issue either.

Let's wait and see if this becomes an issue. Thanks either way! 💪

PR #64, opened 2019-08-28
Ability to specify status code

### LINKED ISSUE #56: Improve developer experience for handling server errors
There's a lot of confusion around error handling at the moment. When a server error happens, Laravel returns a non-Inertia response, which means Inertia renders the response in a modal.

People often ask for a way to disable the modal behaviour. This is not the idiomatic solution to the problem; the correct solution is to ensure your server _always_ returns an Inertia response.

In Laravel, this can be achieved by modifying the [exception handler](https://github.com/laravel/laravel/blob/bb433725483803a27f21d3b21317072610bc3e9c/app/Exceptions/Handler.php#L40-L50).

A basic example:

```php
public function render($request, Exception $exception)
{
    if ($exception instanceof ModelNotFoundException) {
        return Inertia::render('Errors/NotFound');
    }

    return parent::render($request, $exception);
}
```

This is a tedious job; you need to ensure you've caught all default Laravel cases, and it's a lot of boilerplate code to add to every project. 

Laravel's default bahaviour is to render a Blade view based on the error's response code, like `errors.404`, `errors.500` etc. The inertia-laravel package could mimic this method.

```php
public function render($request, Exception $exception)
{
    return Inertia::renderException($request, $exception);
}
```

We could also ship a "check" for applications with multiple parts, including non-Inertia parts like an API or separate homepage.

```php
public function render($request, Exception $exception)
{
    if (Inertia::isInertiaRequest($request)) {
        return Inertia::renderException($request, $exception);
    }

    return parent::render($request, $exception);
}
```

This would return an response with a standardized component name, for example `Inertia::render('Errors/NotFound')`. Note that JavaScript functions aren't allowed to start with a number, so we'd need to either use explicit names (`NotFound`), or a prefix (`Error404`).

What needs to be done?

- First of all, this idea needs to be validated. Looking for input!
- We'd need to create a mapping of Laravel exceptions to component names
- There needs to be a way to specify the response code in `Inertia::render`

hivokas:
# Description

This PR adds an ability to specify the status code.

# Usage

With this improvement you can use inertia components inside the exception handler with an ability to specify the correct status code (e.g. for 404 page).

Partly resolves #56


src/Response.php
```diff
@@ -24,6 +24,13 @@ public function __construct($component, $props, $rootView = 'app', $version = nu
         $this->version = $version;
     }
 
+    public function withStatusCode($statusCode)
```
sebastiandedeyne: Laravel has a `setStatusCode` method on the response factory, maybe we should use the same method name?

hivokas: Yeah, why not?

reinink:
Hey @hivokas, could you update the PR to include a description of what you're trying to accomplish here? Like, I get that you want to add status codes to response...but I'm curious why?

hivokas:
I've updated the description.

src/Response.php
hivokas: Done

reinink:
@hivokas I merged another PR (#44), which has some implications on this PR. Do you mind updating? 🙏

hivokas:
I don't have an access to my laptop now. So maybe in a week. Ok?

reinink:
@hivokas For sure, take your time. 👍

reinink:
Hey @hivokas! Sorry for the delay. Curious, why did you close this?

hivokas:
@reinink is it still relevant?

reinink:
I'm not sure, to be honest. I haven't had the need for this functionality. Nonetheless, this issue is here if we ever want to revisit. 👍

PR #65, opened 2019-09-01
Remove registering Inertia middleware from readme

aryehraber:
Noticed that this step is no longer needed thanks to [this PR](https://github.com/inertiajs/inertia-laravel/pull/61). Love how easy it is to get started with Inertia 💪


sebastiandedeyne:
Thanks!

PR #68, opened 2019-09-04
Remove array_only in favour of facade array helper

rodrigore:
Hi, I saw on twitter the partial reload feature and it didn't work on my projects because they are on laravel 5.8 and 6.0.
https://twitter.com/inertiajs/status/1160182258478788608

So I noticed that the reason was that the code use 'array_only' helper that it was deprecated since laravel 5.8.

This pull use facade array helper in order to use laravel 5.8 and laravel 6.0

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -3,6 +3,7 @@
 namespace Inertia;
 
 use Closure;
+use Illuminate\Support\Arr;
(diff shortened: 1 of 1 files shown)


reinink:
Perfect, thanks!

Issue #66, opened 2019-09-02
[Bug] Inertia Route macro does not work with Route Parameters


marvinrabe:
In web.php I have:

    Route::inertia('{any}', 'MyPage');

When calling `http://foo.test/` this works fine. MyPage will be loaded. If I call `http://foo.test/bar` an exception is raised:

```
ErrorException thrown with message "array_merge(): Expected parameter 2 to be an array, string given"

Stacktrace:
#59 ErrorException in /Users/marvin/Sites/laravel/vendor/inertiajs/inertia-laravel/src/ResponseFactory.php:53
#58 array_merge in /Users/marvin/Sites/laravel/vendor/inertiajs/inertia-laravel/src/ResponseFactory.php:53
#57 Inertia\ResponseFactory:render in /Users/marvin/Sites/laravel/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php:237
#56 Illuminate\Support\Facades\Facade:__callStatic in /Users/marvin/Sites/laravel/vendor/inertiajs/inertia-laravel/src/Controller.php:9
#55 Inertia\Controller:__invoke in /Users/marvin/Sites/laravel/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php:48
#54 Illuminate\Routing\ControllerDispatcher:dispatch in /Users/marvin/Sites/laravel/vendor/laravel/framework/src/Illuminate/Routing/Route.php:219
```

## Debugging

When I dump the arguments of the render function in `inertiajs/inertia-laravel/src/ResponseFactory.php`. I get the following:

`$component` is `bar` and `$props` is `MyPage`.

It seems like there are Issues with the route bindings in the Inertia Controller when the route contains a route parameter. When I change the Inertia Controller to be like this: 

```
public function __invoke($component, $props, $x)
    {
        dd($component, $props, $x);
        return Inertia::render($component, $props);
    }
```

This will dump `$component` is `bar` and `$props` is `MyPage` and `$x` is an empty array.

So the `any` parameter in the route is mapped to `$component`. Which is obviously wrong.

## Workaround

A possible solution for this might be the following:

```
public function __invoke()
    {
        return Inertia::render(request('component'), request('props'));
    }
```

sebastiandedeyne:
This looks like a bug to me. @peterfox any chance you could do a follow up PR that fixes this? 😇

peterfox:
@sebastiandedeyne sure I'll try and fit in some time to take a look

[...]

PR #69, opened 2019-09-04
Fixes wild card URLs used with route macro

peterfox:
Resolves Issue #66

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -2,10 +2,15 @@
 
 namespace Inertia;
 
+use Illuminate\Http\Request;
+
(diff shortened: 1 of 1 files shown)


reinink:
@peterfox Just realizing now that this change is now breaking the `ControllerTest`. Are you able to pull down master and fix this?

Previously the `Controller@__invoke` method took a `$component` and `$props`. Now it takes a `Request`. I'm not really sure exactly how this feature is supposed to work...so you help here would be appreciated.

Once you're done that, I'll tag a new release.

Issue #27, opened 2019-05-22
support for API Resources?


ceejayoz:
Is support for Eloquent's [API Resources](https://laravel.com/docs/5.8/eloquent-resources) planned? We have models that a) have properties only available after a certain date and b) other properties only available to admin users.

With an API resource, we can just do stuff like:

```
$return = [
    'id' => $this->id,
    'name' => $this->id,
];

if(auth()->user()->isAdmin) {
    $return['answer'] = 'foo';
}

return $return;
```

Being able to do this would be spectacular:

    return Inertia::render('Model')
        ->with('model', new ModelResource($model);

and this:

    return Inertia::render('List')
        ->with('models', new ModelResource::collection(Model::all());

ceejayoz:
Aaaaaand I should've tried it before writing this up.

adriandmitroca:
Hi @ceejayoz!

Have you somehow managed to how pass pagination data alongside models?

ceejayoz:
@adriandmitrica I have not. It seems to be added to the resource via the toResponse function, which Inertia doesn’t trigger. 

Trying to completely confirm before I embarrass myself again with another issue. 🤪

ceejayoz:
@adriandmitroca I opened #28 about this.

adriandmitroca:
Good one!

adriandmitroca:
This is weird, in my case `$this->links()` returns null. Are you returning this within `toArray` method?

reinink:
I've added support for "Responsable" props: 2b966ac27964534ee3bbe585b0ea0fff8babb53a

I've tested this with both single resource responses and collection resource responses, and it works great. 👍

Issue #28, opened 2019-05-24
API resources don't include pagination data when passed to Inertia render


ceejayoz:
Pagination data in API resources appears to be handled by the [`toResponse` function](https://github.com/laravel/framework/blob/bf7076f9a325a65770aa802d6014993422eb8f6c/src/Illuminate/Http/Resources/Json/PaginatedResourceResponse.php#L15) of `Illuminate\Http\Resources\Json\PaginatedResourceResponse`.

That means this:

    return Inertia::render('Foo', [
        'foo' => FooResource::collection(Foo::paginate()),
    ]);

gets the correctly transformed collection, but it's missing the `links` (and other) metadata.

I'm able to work around it by doing this:

    return Inertia::render('Foo', [
        'foo' => new FooCollection(Foo::paginate()),
    ]);

and having FooCollection do this:

    return [
        'data' => FooResource::collection($this->collection),
        'links' => $this->links()
    ];

but it's not as elegant as the default behavior in Laravel.

adriandmitroca:
This is weird, in my case `$this->links()` returns null. Are you returning this within `toArray` method?

ceejayoz:
Here's both Resource classes, @adriandmitroca.

```<?php

namespace App\Http\Resources;

use Illuminate\Http\Resources\Json\JsonResource;

class FooResource extends JsonResource
{
    /**
     * Transform the resource into an array.
     *
     * @param  \Illuminate\Http\Request  $request
     * @return array
     */
    public function toArray($request)
    {
        return [
            'id' => $this->id,
            'name' => $this->name,
        ];
    }
}```


```
<?php

namespace App\Http\Resources;

use Illuminate\Http\Resources\Json\ResourceCollection;

class FooCollection extends ResourceCollection
{
    /**
     * Transform the resource collection into an array.
     *
     * @param  \Illuminate\Http\Request  $request
     * @return array
     */
    public function toArray($request)
    {
        return [
            'data' => FooResource::collection($this->collection),
            'links' => $this->links()
        ];
    }
}
```

Make sure you're extending `ResourceCollection` instead of `JsonResource` in the collection one.

adriandmitroca:
Yeah, that's basically exactly what I have. Is it working for you?

adriandmitroca:
I see the issue - Laravel returns huge object of `Illuminate\Support\HtmlString` with entire HTML markup for default render method in this situation rather than actual array with pagination data.

:/

reinink:
[...]

PR #70, opened 2019-09-06
Support API Resources

winter-ice:
resolve https://github.com/inertiajs/inertia-laravel/issues/27 https://github.com/inertiajs/inertia-laravel/issues/28
Use like:
```
$user = UserResource::make(User::find(1));
return Inertia::render('Users/Edit', $user)
```
```
$users = UserResource::collection(User::paginate());
return Inertia::render('Users/Index', $users)
```
or
```
$users = UserCollection::make(User::paginate());
return Inertia::render('Users/Index', $users);
```

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -3,6 +3,7 @@
 namespace Inertia;
 
 use Closure;
+use Illuminate\Http\Resources\Json\JsonResource;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\App;
 use Illuminate\Contracts\Support\Arrayable;
@@ -48,6 +49,10 @@ class ResponseFactory
 
     public function render($component, $props = [])
     {
+        if ($props instanceof JsonResource) {
+            $props = $props->response()->getData(true);
+        }
+
         if ($props instanceof Arrayable) {
             $props = $props->toArray();
         }


reinink:
Hmm, I thought this worked already? I've used API resources. They automatically get converted to JSON, so this shouldn't be needed.

Plus, we also added `Arrayable` support recently (although it's not tagged yet): #39

I'll tag that now. Can you try this again, on the latest version? I suspect it will work without any changes needed to the library.

winter-ice:
1. Passed to render function is an instance of `JsonResource`,Not instance of `Arrayable`.So I think it's not work.
2. Before response,The data not be fully handled.(such as `meta`,`link`).

forgive my poor English 😄

winter-ice:
And I have test many usage.Here is some examples:
```
$user = User::find(1);
$props = UserResource::make($user);

return Inertia::render('Users/Edit', $props)
```
```
$user = User::find(1);
$props = UserResource::make($user)->additional(['foo'=>'bar']);

return Inertia::render('Users/Edit', $props)
```
And the `with` method in your resource class is works well.

Notice that if you want to append other data to a collection.
there are two ways:
1. use `additional`
2. defined `with` method in your resource collection.

defined in a resource and use `UserResource::collection` is not works

sebastiandedeyne:
> Before response,The data not be fully handled.(such as meta,link).

This is true, and a super annoying thing with Laravel resources. When calling `toArray` you lose meta and pagination data.

It's debatable whether Inertia should cast this for the user though, since some people might expect to only get the data when passing down an Eloquent resource collection.

I personally would be very happy with this change, but I don't know what others are doing with resources; it might be to opinionated.

.gitignore
```diff
@@ -1,4 +1,5 @@
 /vendor
+.idea
```
m1guelpf: This change seems unrelated to the PR

winter-ice: Yes,It's unrelated to the PR,I just get used to add this.

sebastiandedeyne: Editor settings should be globally ignored :) https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

winter-ice: Thanks,I don't know that.

m1guelpf:
```suggestion
```

thoresuenert:
This PR will save time and a Collection class if you go the Laravel resource route.

In our team the resource is a good api for transforming data, I like it a lot.

src/ResponseFactory.php
```diff
@@ -48,6 +49,10 @@ public function getVersion()
 
     public function render($component, $props = [])
     {
+        if ($props instanceof JsonResource) {
```
grantholle: So in my case, I would be passing an array of props, with some being an instance of `JsonResource`.

It would need to iterate each of the props to resolve them as well. It may be worth rethinking this a little...

lexdewilligen:
@reinink Any updates on this? I need the resource meta, so I'd love to see one of the solutions merged..

reinink:
So, I've added support for "Responsable" props: 2b966ac27964534ee3bbe585b0ea0fff8babb53a

As you can see, my approach here is almost identical to this PR. However, instead of checking for a `JsonResource` object, I'm checking for the more generic `Responsable` object. This makes it useful in other situations as well.

I've tested this with both single resource responses and collection resource responses, and it works great.

Thanks, either way, for submitting this PR @winter-ice. 👍

PR #72, opened 2019-09-10
Add config file

manueldublanc:
In this PR I added a config file to the package. It contains the name of the root template and allows the developer to overwrite the default value with his own preference in a config file (beside the option with `Inertia::setRootView()`).

I also added a short description to the readme with the artisan `vendor:publish` command to copy the config file into the project.

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -9,9 +9,14 @@ use Illuminate\Contracts\Support\Arrayable;
 
 class ResponseFactory
 {
-    protected $rootView = 'app';
+    protected $rootView;
     protected $sharedProps = [];
     protected $version = null;
+
+    public function __construct()
+    {
+        $this->rootView = config('inertia.root_template');
+    }
 
     public function setRootView($name)
     {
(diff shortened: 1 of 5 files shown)


src/ResponseFactory.php
```diff
@@ -9,10 +9,15 @@
...
 class ResponseFactory
 {
-    protected $rootView = 'app';
+    protected $rootView;
     protected $sharedProps = [];
     protected $version = null;
 
+    public function __construct()
+    {
+        $this->rootView = config('inertia.root_template');
```
RSpeekenbrink: Maybe add a default value of ```'app'``` in the config function for when the config does not exist. Like:

```
$this->rootView = config('inertia.root_template', 'app');
```

RSpeekenbrink:
The changes in Readme.md are irrelevant now the docs have been moved to the https://github.com/inertiajs/inertiajs.com repo. Maybe add the docs there instead?

reinink:
@manueldublanc Thanks so much for this contribution, but I don't think I want to introduce a config file yet, if we can avoid it.

PR #75, opened 2019-09-26
Bind ResponseFactory as a singleton

### LINKED ISSUE #32: Test response view data doesn't contain shared props
I'm trying to test a route that should pass the auth user as a shared prop data, along with others(app.name...etc). I am only seeing the prop data defined in the controller.

Anyone run into this issue? Any suggestions... @reinink 

btw, all shared prop data are being passed down to the Vue page component successfully and the json_encoded version can be seen in page view source. I just can't test it.

### LINKED ISSUE #73: HTTP testing fails when a custom root view is set: `View [app] not found`
I think this is happening because the `Inertia\ResponseFactory` class is being instantiated twice during the course of the test, but the rootView property is only reset the first time. We could sort this out by injecting the root view as a nullable constructor parameter and then binding the value of the `$rootView` primitive requested by the Responsable class in the `setRootView()` method.

If this solution makes sense I can put together a PR.

rojtjo:
Fixes #73 
Fixes #32

More info in this comment https://github.com/inertiajs/inertia-laravel/issues/73#issuecomment-535604070

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -16,6 +16,11 @@ class ServiceProvider extends BaseServiceProvider
         $this->registerMiddleware();
     }
 
+    public function register()
+    {
+        $this->app->singleton(ResponseFactory::class);
+    }
+
     protected function registerBladeDirective()
     {
         Blade::directive('inertia', function () {


reinink:
Interesting, thanks for submitting this @rojtjo. Curious where you got the solution? Wondering if there is any precedent you point to for this solution?

rojtjo:
The  first thing I thought when I saw the error was "Ahh, missing binding!", but then it left me wondering how `share()` was actually working without the binding. So I ended up reading through the Facade and noticed that is was actually caching the resolved instances (which in turn makes them act like singletons until cleared).

johanvanhelden:
Could this get merged? I'm currently unable to write tests for my Inertia app due to this issue.

mattharcourt:
This also appears to solve #32

bastien-phi:
@mattharcourt You can bind the singleton by yourself in `register` of your `AppServiceProvider` for example. 
@reinink any idea when you will have the time to merge this ? Thanks for your work and your time, Inertia is just awesome ! :fire:

reinink:
Thanks for this @rojtjo! 👍

AndrewFeeney:
Thanks @rojtjo and @reinink 🙌

PR #80, opened 2019-10-29
isInertiaRequest function

RSpeekenbrink:
For some specific use cases I would like to check if the given request is an Inertia Request. I can check for the Inertia header in the request object but to keep the header name 'dynamic' and changeable later without breaking the package user's code it would be better to create a function in the Facade that checks if a given request is an InertiaRequest. This PR adds the ```isInertiaRequest``` function to the Inertia facade (ResponseFactory). It can be used as follows:

```
class SomeController
{
    {
        if (Inertia::isInertiaRequest($request)) {
            // Inertia Request
        }
    }
}
```

### DIFF
--- a/src/Inertia.php
+++ b/src/Inertia.php
@@ -5,6 +5,7 @@ namespace Inertia;
 use Illuminate\Support\Facades\Facade;
 
 /**
+ * @method static bool isInertiaRequest($request)
  * @method static void setRootView($name)
  * @method static void share($key, $value = null)
  * @method static array getShared($key = null)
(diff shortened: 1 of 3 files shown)


reinink:
I like it, but I think there is a better approach. What if we simply added a macro the Laravel `Request` object in the Inertia service provider. Then you could just do this:

```php
if ($request->inertia()) {
    // do thing
}
```

This is similar to the `pjax` check: https://github.com/laravel/framework/blob/66eb5d01609f1021baed99a3f5d9d16bec6dcfcb/src/Illuminate/Http/Request.php#L248-L251

I'm going to add this. 👍

PR #91, opened 2020-01-02
Make response factory macroable

### LINKED ISSUE #84: Make the Inertia macroable
Hi there nice work 🎉
I'm developing my admin app using inertia and I got some Resource classes on my front-end so all my controllers pass a `resource` parameter to pages like this:
```php
Inertia::render('Resource/Index',[
    'resource' => 'BlogPost',
    // ...
])
```
so it's good unless i have more than one methods on my resource controller ( mostly all CRUD methods ). so i prefer `share` resource parameter in `__constructor` method instead of repeating it at every method like :
```php
Class BlogPostController extends Controller{
    __constructor(){
        Inertia::share('resource','BlogPost');
    }
    //...
}
```

So i though it's cleaner to make Inertia macroable so it would be possible do something like:
```php
Class BlogPostController extends Controller{
    __constructor(){
        Inertia::resource('BlogPost');
    }
    //...
}
```

reinink:
Closes #84

This makes the `ResponseFactory`, and subsequently the `Inertia` facade, macroable.

```php
Inertia::macro('foo', function () {
    return 'bar';
});
```

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -5,10 +5,13 @@ namespace Inertia;
 use Closure;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\App;
+use Illuminate\Support\Traits\Macroable;
 use Illuminate\Contracts\Support\Arrayable;
 
 class ResponseFactory
 {
+    use Macroable;
+
     protected $rootView = 'app';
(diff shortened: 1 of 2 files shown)


lexdewilligen:
@reinink Awesome to see inertia move forward again!

PR #112, opened 2020-04-29
Add `Arrayable` docblock for static analysis

brendt:
`$instance->render($component, $props);` can deal with arrays and arrayable objects, though the docblock on the `inertia` helper function only specifies `array` as a possible type. This causes static analysis tools to complain when you pass in an arrayable.


claudiodekker (approved):

reinink:
Thanks @brendt!

PR #116, opened 2020-06-21
Allow a custom root view per response

robinvdvleuten:
I have an application where the layout / assets differs on different controller groups. This PR allows to set a different layout within the controller action while returning the Inertia response.

```php
return Inertia::render('Event/Show', [
    'event' => $event->only(
        'id',
        'title',
        'start_date',
        'description'
    ),
])->withRootView('custom');
```

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -37,6 +37,13 @@ class Response implements Responsable
         return $this;
     }
 
+    public function withRootView($rootView)
+    {
+        $this->rootView = $rootView;
+
+        return $this;
+    }
+
     public function withViewData($key, $value = null)
     {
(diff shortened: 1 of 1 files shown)


reinink:
So, my concern here is that this will make people think that this will somehow change the root view when making Inertia requests (since it looks like that). I'd recommend just create a macro for this, if this is something you want to do.

Also, we just released [v0.3.2](https://github.com/inertiajs/inertia-laravel/releases/tag/v0.3.2), which adds the ability to set the root template via the new Inertia middleware. I'd recommend conditionally setting this custom root template via the new `rootView()` method. 👍 

I hope that helps!

PR #119, opened 2020-06-25
Make Inertia::share support Arrayable classes

hailwood:
Right now we can do `Inertia::render('Page', new SomeArrayableClass());` and it works.

Unfortunately `Inertia::share` doesn't work the same and we need to do `Inertia::share((new SomeArrayableClass())->toArray())`.

This PR allows `share` to automatically transform an Arrayable class.


claudiodekker (approved):

reinink:
Thanks @hailwood!

PR #120, opened 2020-06-30
Added test for partial reload

Juhlinus:


tests/ResponseTest.php
```diff
@@ -160,4 +160,27 @@ public function toArray()
...
+        $user = (object) ['name' => 'Jonathan'];
+        $response = new Response('User/Edit', ['user' => $user, 'partial' => 'partial-data'], 'app', '123');
+        $response = $response->toResponse($request);
+        $page = $response->getData();
+
+        $props = get_object_vars($page->props);
+
+        $this->assertInstanceOf(JsonResponse::class, $response);
+        $this->assertSame('User/Edit', $page->component);
+        $this->assertFalse(isset($props['name']));
```
victorlap, line 180: Shouldn’t you check for the user key here?

Juhlinus: The user key is not part of the partial data, the only key present is the `partial` key. See Line 169

victorlap: Yeah, that's why you want to check it is not there right? The `name` prop is within the `user` right?

```diff
- $this->assertFalse(isset($props['name']));
+ $this->assertFalse(isset($props['user']));

Juhlinus: Ah! I see what you mean now. Good catch!

I'll update this after work today. 😃

reinink:
Looks good @Juhlinus with that one change. 💪

Juhlinus:
@reinink done!

reinink:
Thanks buddy! 💪

PR #105, opened 2020-03-09
Inertia fake / Inertia test helpers

ChristianPavilonis:
I know @reinink has helpers coming, but I thought I'd make a PR if anyone wants to try out my fork and use some helpers now.

### To try this fork out in an existing project update composer.json.
```
  "require": {
        .....
        "inertiajs/inertia-laravel": "dev-master", // change to dev-master
    },
    "repositories": [
        { // use forked repo.
            "type" : "vcs",
            "url" : "https://github.com/ChristianPavilonis/inertia-laravel.git"
        }
    ],
```
Run `composer update`

### Usage:
In your test call the fake
```
Inertia::fake();
```

After that, you can use the assertions 

```
Inertia::assertHasProps(['prop1', 'prop2]);

Inertia::assertHasProp('prop_name');

Inertia::assertPropsEqual($expectedProps);

Inertia::assertPropEquals($key, $expectedValue);

Inertia::assertComponentIs('ComponentName');

// You can also access the props
Inertia::getAllProps();

Inertia::getProp($key);
```

PR #124, opened 2020-07-17
(Yet Another) Assertions implementation

claudiodekker:
## Motivation
Like others, I've been in search of a way to test the backend outputs towards an Inertia view. While Macros are the obvious solution that others have also already discovered & suggested via PR, there were some things there that just didn't feel 'right' to me.

For example, while I do believe https://github.com/inertiajs/inertia-laravel/pull/51 to be a great implementation, I wasn't a big fan of the idea of having to swap out the `TestCase` class in our own projects. Similarly, while the suggestions made in https://github.com/inertiajs/inertia-laravel/pull/105 aren't bad either, but having to run `Inertia::fake()` was kind of a dealbreaker to me.

So, here's my attempt.

## Usage

This PR suggests a couple of assertion helpers that are available out of the box once you update to whatever version this PR ships in. To use, simply chain them on your `TestResponse` responses. 

Screenshot 2020-09-02 at 19 44 39

## Available Methods
The methods made available by this PR closely reflect those available in Laravel itself.
(_In fact, I just got a notification that [a few changes I suggested were just merged into the framework](https://github.com/laravel/framework/pull/33566)_)

---

Assert whether the given page is an Inertia-rendered view
```php
$response->assertInertia();

// or, also check whether the page is a specific component
$response->assertInertia('example');

// or, also check whether all of the given props match
$response->assertInertia('example', [
    'foo' => 'bar'
]);
```

Return all the available Inertia props for the page
``` php
$response->inertiaProps(); 
```

Assert whether the Inertia-rendered view has a specific property set
```php
$response->assertInertiaHas('key');

// or, against deeply nested values
$response->assertInertiaHas('deeply.nested.key');
```

Apart from checking whether the property is set, the same method can be used to assert that the values match
```php
$response->assertInertiaHas('key', 'matches-this-value');

// or, for deeply nested values
$response->assertInertiaHas('deeply.nested.key', 'also-match-against-this-value');
```

It's also possible to assert directly against a Laravel Model (or any other `Arrayable` or `Responsable` class)
```php
$user = UserFactory::new()->create(['name' => 'John Doe']);

// ... (Make HTTP request etc.)

$response->assertInertiaHas('user', $user);
$response->assertInertiaHas('deeply.nested.user', $user);
```

It's also possible to check against a closure
```php
$response->assertInertiaHas('foo', function ($value) {
    return $value === 'bar';
});

// or, again, for deeply nested values
$response->assertInertiaHas('deeply.nested.foo', function ($value) {
    return $value === 'bar';
});
```

Next, you can also check against a whole array of properties. It'll simply loop over them using the `assertInertiaHas` method described above:
```php
$response->assertInertiaHasAll([
    'foo',
    'bar.baz',
    'another.nested.key' => 'example-value'
]);
```

Finally, you can assert that a property was not set:
```php
$response->assertInertiaMissing('key');

// or, for deeply nested values
$response->assertInertiaMissing('deeply.nested.key');
```

## Details & (Breaking) Changes
This PR bootstraps the Macros directly onto the `TestResponse` from the ServiceProvider, and only does so if we're running in a testing environment. 

I've also made it compatible with Laravel 6 and earlier, as that's where the `TestResponse` namespace was changed from `Illuminate\Foundation\Testing\TestResponse` to `Illuminate\Testing\TestResponse`.

Finally, I've changed the package constraints of this repo to be only compatible with Laravel 5.4 and up, because [Laravel 5.3 depends upon `phpunit/phpunit: ~5.4`](https://github.com/laravel/framework/blob/5.3/composer.json#L79), which was when the non-namespaced PHPUnit classes (e.g. [`PHPUnit_Framework_Assert`](https://github.com/sebastianbergmann/phpunit/blob/5.4.0/src/Framework/Assert.php#L16)) were still used. ([Laravel 5.4 changed this](https://github.com/laravel/framework/blob/5.4/composer.json#L78)).

While this might seem like a breaking change, I'm fairly confident it isn't, as on those versions other things in this package already likely wouldn't have worked 😅

Let me know if this is something you're happy with, or whether you want me to tweak certain things.
(Or, alternatively, there's no hard feelings if you decide to just decline this PR altogether, so feel free to do so..)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,8 +24,12 @@
             "Inertia\\Tests\\": "tests/"
         }
     },
+    "require": {
+        "laravel/framework": "^5.4|^6.0|^7.0"
+    },
     "require-dev": {
-        "orchestra/testbench": "~3.0"
+        "orchestra/testbench": "^3.4|^4.0|^5.0",
+        "phpunit/phpunit": "~5.7|~6.0|^7.0|^8.0|^9.0"
     },
     "extra": {
         "laravel": {
--- a/ide_helpers.php
+++ b/ide_helpers.php
@@ -1,0 +1,41 @@
+<?php
+
+/** @noinspection PhpUndefinedClassInspection */
+/** @noinspection PhpFullyQualifiedNameUsageInspection */
+/** @noinspection PhpUnusedAliasInspection */
+
+namespace Illuminate\Testing {
+
+    /**
+     * @see \Inertia\Assertions
+     *
+     * @method self assertInertia()
+     * @method self assertInertiaComponent($name)
+     * @method self assertInertiaHas($key, $value = null)
+     * @method self assertInertiaHasAll(array $bindings)
+     * @method self assertInertiaMissing($key)
+     * @method array inertiaProps()
+     */
+    class TestResponse
+    {
+        //
+    }
+}
+
+namespace Illuminate\Foundation\Testing {
+
+    /**
+     * @see \Inertia\Assertions
+     *
+     * @method self assertInertia()
+     * @method self assertInertiaComponent($name)
+     * @method self assertInertiaHas($key, $value = null)
+     * @method self assertInertiaHasAll(array $bindings)
+     * @method self assertInertiaMissing($key)
+     * @method array inertiaProps()
+     */
+    class TestResponse
+    {
+        //
+    }
+}
--- a/ide_helpers.php
+++ b/ide_helpers.php
@@ -9,8 +9,7 @@ namespace Illuminate\Testing {
     /**
      * @see \Inertia\Assertions
      *
-     * @method self assertInertia()
-     * @method self assertInertiaComponent($name)
+     * @method self assertInertia($component = null, $props = [])
      * @method self assertInertiaHas($key, $value = null)
      * @method self assertInertiaHasAll(array $bindings)
      * @method self assertInertiaMissing($key)
@@ -27,8 +26,7 @@ namespace Illuminate\Foundation\Testing {
     /**
      * @see \Inertia\Assertions
      *
-     * @method self assertInertia()
-     * @method self assertInertiaComponent($name)
+     * @method self assertInertia($component = null, $props = [])
      * @method self assertInertiaHas($key, $value = null)
      * @method self assertInertiaHasAll(array $bindings)
      * @method self assertInertiaMissing($key)
--- a/src/Assertions.php
+++ b/src/Assertions.php
@@ -3,16 +3,16 @@
 namespace Inertia;
 
 use Closure;
-use Illuminate\Support\Arr;
-use PHPUnit\Framework\Assert as PHPUnit;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Support\Responsable;
+use Illuminate\Support\Arr;
+use PHPUnit\Framework\Assert as PHPUnit;
 
 class Assertions
 {
     public function assertInertia()
     {
-        return function () {
+        return function ($component = null, $props = []) {
             $this->assertViewHas('page');
 
             tap($this->viewData('page'), function ($view) {
@@ -21,6 +21,12 @@ class Assertions
                 PHPUnit::assertArrayHasKey('url', $view);
                 PHPUnit::assertArrayHasKey('version', $view);
             });
+
+            if (! is_null($component)) {
+                PHPUnit::assertEquals($component, $this->viewData('page')['component']);
+            }
+
+            $this->assertInertiaHasAll($props);
 
             return $this;
         };
@@ -32,17 +38,6 @@ class Assertions
             $this->assertInertia();
 
             return $this->viewData('page')['props'];
-        };
-    }
-
-    public function assertInertiaComponent()
-    {
-        return function ($name) {
-            $this->assertInertia();
-
-            PHPUnit::assertEquals($name, $this->viewData('page')['component']);
-
-            return $this;
         };
     }
 
--- a/src/Assertions.php
+++ b/src/Assertions.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace Inertia;
+
+use Closure;
+use Illuminate\Support\Arr;
+use Illuminate\Database\Eloquent\Model;
+use PHPUnit\Framework\Assert as PHPUnit;
+
+class Assertions
+{
+    public function assertInertia()
+    {
+        return function () {
+            $this->assertViewHas('page');
+
+            tap($this->viewData('page'), function ($view) {
+                PHPUnit::assertArrayHasKey('component', $view);
+                PHPUnit::assertArrayHasKey('props', $view);
+                PHPUnit::assertArrayHasKey('url', $view);
+                PHPUnit::assertArrayHasKey('version', $view);
+            });
+
+            return $this;
+        };
+    }
+
+    public function inertiaProps()
+    {
+        return function () {
+            $this->assertInertia();
+
+            return $this->viewData('page')['props'];
+        };
+    }
+
+    public function assertInertiaComponent()
+    {
+        return function ($name) {
+            $this->assertInertia();
+
+            PHPUnit::assertEquals($name, $this->viewData('page')['component']);
+
+            return $this;
+        };
+    }
+
+    public function assertInertiaHas()
+    {
+        return function ($key, $value = null) {
+            if (is_array($key)) {
+                return $this->assertInertiaHasAll($key);
+            }
+
+            if (is_null($value)) {
+                PHPUnit::assertArrayHasKey($key, $this->inertiaProps());
+            } elseif ($value instanceof Closure) {
+                PHPUnit::assertTrue($value(Arr::get($this->inertiaProps(), $key)));
+            } elseif ($value instanceof Model) {
+                PHPUnit::assertEquals($value->toArray(), Arr::get($this->inertiaProps(), $key));
+            } else {
+                PHPUnit::assertEquals($value, Arr::get($this->inertiaProps(), $key));
+            }
+
+            return $this;
+        };
+    }
+
+    // TODO: Test.
+    public function assertInertiaHasAll()
+    {
+        return function (array $bindings) {
+            foreach ($bindings as $key => $value) {
+                if (is_int($key)) {
+                    $this->assertInertiaHas($value);
+                } else {
+                    $this->assertInertiaHas($key, $value);
+                }
+            }
+
+            return $this;
+        };
+    }
+
+    // TODO: Test.
+    public function assertInertiaMissing()
+    {
+        return function ($key) {
+            $this->assertInertia();
+
+            PHPUnit::assertArrayNotHasKey($key, $this->inertiaProps());
+
+            return $this;
+        };
+    }
+}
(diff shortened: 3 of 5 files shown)


johanvanhelden:
Thanks for the effort! I love the implementation. 

How would you handle testing against [resources and collections](https://laravel.com/docs/7.x/eloquent-resources)?
In my projects only pass resources to my Inertia pages to prevent unwanted attributes from bleeding to the front end.

So would something like this be possible?

```php
$projects = factory(Project::class, 5)->states(['public', 'published'])->create();

// snip

$response->assertPropValue('projects', ProjectResource::collection($projects->sortByDesc('published_at')));
```

I've tried myself to make a macro to handle this, but it is quite a pain. Especially when utilizing nested resources, like this:

```php
<?php

// snip

/** @extends JsonResource<\App\Models\Project> */
class ProjectResource extends JsonResource
{
    public function toArray($request)
    {
        return [
            'uuid' => $this->uuid,
            'name' => $this->name,
            
            'owner'    => new UserResource($this->owner),
            'images'   => ImageResource::collection($this->images),

            'urls' => [
                'show' => 'demo.com',
            ],
        ];
    }
}

```

I just can not get a proper value assert against a resource collection like that.

claudiodekker:
Interesting use case, I hadn't considered that one.. I'll take a look at it when have some time!

claudiodekker:
Alright, so, after letting your question brew for a bit, I realized that normally my answer would be that "_[you're holding it wrong](https://knowyourmeme.com/memes/events/iphone-4-death-grip)_", because [Eloquent API Resources](https://laravel.com/docs/7.x/eloquent-resources) are meant for exactly that: APIs. And since Inertia doesn't _really_ have an API, I would instead recomment to look into [using Eloquent's built-in `$hidden` property to hide attributes from the view](https://laravel.com/docs/7.x/eloquent-serialization#hiding-attributes-from-json)

However, at the same time, @reinink did support it as per 2b966ac27964534ee3bbe585b0ea0fff8babb53a, so I've still gone ahead and implemented it. Enjoy!

johanvanhelden:
Thank you @claudiodekker ! I will test it out tomorrow (hopefully). 

The reason why I love API Resources so much is because you can create different resources for the same model for different situations, conditionally add data, etc. For example, you can create a ProjectIndexResource for a list of projects and a ProjectResource that is used everywhere else.  And if your project has an API, you can add them there as well. Win-win.

It's just extremely flexible compared to the the all-or-nothing `$hidden`.

johanvanhelden:
Reporting back after testing. It seems to work great for single-depth resources. But if I start nesting resources, I can not really get the assertions to work. Could you tell me if I am doing something wrong @claudiodekker ?

I'll try to be as detailed as possible.

This is what it looks like in the Vue inspector:
Screenshot from 2020-07-25 11-57-09


This is what the UserResource looks like:
```php
namespace App\Http\Resources;

use Illuminate\Http\Resources\Json\JsonResource;

class UserResource extends JsonResource
{
    public function toArray($request)
    {
        return [
            'name'  => $this->name,
            'email' => $this->email,

            'roles' => RoleResource::collection($this->whenLoaded('roles')),
        ];
    }
}

```

And the RoleResource:
```php
namespace App\Http\Resources;

use Illuminate\Http\Resources\Json\JsonResource;

class RoleResource extends JsonResource
{
    public function toArray($request)
    {
        return [
            'name' => $this->name,
        ];
    }
}

```

Finally, the controller:
```php
class ProfileController extends Controller
{
    public function index()
    {
        return Inertia::render('Profile/Index', [
            'user' => new UserResource(Auth::user()->load('roles')),
        ]);
    }
```

There is a user object with a roles array inside it. So I assumed this would work:
`$response->assertInertiaHas('user.roles', RoleResource::collection($user->roles));`

But I get the following error: ```null does not match expected type "array".```

When I try to dump the roles using `dump($response->inertiaProps()['user']['roles']);` I get:
`Cannot use object of type stdClass as array`

So I think there is still an issue with nested collections.

This is the test I used:
```php
    /** @test */
    public function shows_the_current_user(): void
    {
        $user = factory(User::class)->state('user')->create(['name' => 'Test Name'])->load('roles');

        $response = $this
            ->actingAs($user)
            ->get(route('profile.index'))

            ->assertSee('Test Name');

        $response->assertInertia(); // == OK
        $response->assertInertiaComponent('Profile/Index'); // == OK
        $response->assertInertiaHas('user', new UserResource($user));  // == OK
        $response->assertInertiaHas('user.roles', RoleResource::collection($user->roles)); // == Error
        dump($response->inertiaProps()['user']['roles']); // == Error
    }
```
I also do not really understand why the `user` asserts **does** work and the `user.roles` assert fails. I would have expected the user assert to also fail :thinking:

claudiodekker:
Hi,

So, I looked into this for you. Here's what's happening.

## The problem
The error you're getting (`Error: null does not match expected type "array".`) occurs because of the `JsonResource` which Eloquent API Resources (eventually) extend. If we take a look at how Inertia prepares properties for the front-end, you can see it calls [$prop = $prop->toResponse($request)->getData();](https://github.com/inertiajs/inertia-laravel/blob/e7b95274dc0f6d4148380e2c45ba1c94f77c1c9c/src/Response.php#L65).
I'll guide you through the relevant bits of 'call stack' below:

1. [return (new ResourceResponse($this))->toResponse($request);](https://github.com/laravel/framework/blob/29c64cb6c1d8cc712463c63fcf60c0a46fe3455e/src/Illuminate/Http/Resources/Json/JsonResource.php#L221)
2. [$this->wrap(...)](https://github.com/laravel/framework/blob/29c64cb6c1d8cc712463c63fcf60c0a46fe3455e/src/Illuminate/Http/Resources/Json/ResourceResponse.php#L38)
3. [$data = [$this->wrapper() => $data];](https://github.com/laravel/framework/blob/29c64cb6c1d8cc712463c63fcf60c0a46fe3455e/src/Illuminate/Http/Resources/Json/ResourceResponse.php#L66)
4. [return get_class($this->resource)::$wrap;](https://github.com/laravel/framework/blob/29c64cb6c1d8cc712463c63fcf60c0a46fe3455e/src/Illuminate/Http/Resources/Json/ResourceResponse.php#L107)

Basically, what happens, is that Laravel looks for the `$wrap` property on your Resource implementation (e.g. `UserResource::class`). However, since it most likely won't have one, it'll instead inherit the property from the `JsonResource` class it extends: [public static $wrap = 'data';](https://github.com/laravel/framework/blob/29c64cb6c1d8cc712463c63fcf60c0a46fe3455e/src/Illuminate/Http/Resources/Json/JsonResource.php#L47)

If we then return/take a step back up the call-stack, you can see that this 'data'-value gets used as the key in a new array, where it's value is the existing instance data. This then again gets returned, and eventually gets send to the browser as a 'Response' object.

However, since Inertia makes this call instead, it then has a `JsonResponse` object, which is somewhat useless to it. 
So, what it does in addition, is calling [`->getData()`](https://github.com/laravel/framework/blob/29c64cb6c1d8cc712463c63fcf60c0a46fe3455e/src/Illuminate/Http/JsonResponse.php#L54) on it, to get the 'json decoded' representation of the response data. This then gets set on the view as a prop.


## Fix number 1 (in _your_ code)
This call:
```php
$response->assertInertiaHas('user.roles', RoleResource::collection($user->roles));
```

Should actually be:
```php
$response->assertInertiaHas('user.data.roles', RoleResource::collection($user->roles));
```

Because the prop you set, is only 'user'. Everything below that gets 'wrapped' in the data key by the Resource:
```php
return Inertia::render('Profile/Index', [
    'user' => new UserResource(Auth::user()->load('roles')),
]);
```

However, doing this, while more correct, still wouldn't make ANY behavioural difference. Here's why:

If we take a look at that `->getData()` call earlier, you can see it plainly makes a `json_decode(...)` call, with the result being an object instead of an associative array (e.g.: `json_decode(..., true)`).

Since this PR (as well as Laravel's `->assertHas`) uses `Arr::get()` and `Arr::has()` to navigate through array properties using dot-notation, and since it cannot (or rather, refuses to) walk through objects, this is where it instead gives back `null`, [which is the default fallback-value in case a key is not found](https://laravel.com/docs/7.x/helpers#method-array-get). 

This also explains the other error you're experiencing: `Cannot use object of type stdClass as array`.

## (Semi-potential) Fix number 2 (in inertia-laravel)

The solution appears to be to **change the way Inertia itself prepares `Responsable` data to the view**, by making [a tiny one-line change](https://github.com/inertiajs/inertia-laravel/blob/e7b95274dc0f6d4148380e2c45ba1c94f77c1c9c/src/Response.php#L65) that effects pretty much nothing else, because further down the line (right now) this stdClass gets re-encoded to JSON anyway:
```diff
- $prop = $prop->toResponse($request)->getData();
+ $prop = $prop->toResponse($request)->getData(true);
```

That way, the `Arr::get()` and `Arr::has()` helpers can walk over the properties, deeply nested and everything (albeit with the `data`-wrapper in between). 

### However, even if we make this change, it only places us in an entirely different problem space:

Due to the serialization that happens, contexts gets lost, so, doing something like this:
```php
$response->assertInertiaHas('example.data.children', $children);
```
will give you the following error:
Screenshot 2020-07-26 at 19 47 48

Alternatively, if we try to match against a child resource that is a `JsonResource`, such as this:
```php
$response->assertInertiaHas('example.data.children', UserResource::collection($children));
```

We'll still get this error, because of the wrapping that I explained earlier:
Screenshot 2020-07-26 at 19 49 57

So, to conclude (at least, for now): I'm not sure if there's a fix that I can apply in the short term. I'll look into it some more, but my best guess is that your best guess (at least, for now) is to **not compare nested resources**.

claudiodekker:
Added IDE testing autocompletion helpers to _this branch_ of laravel-inertia, so if you're using it and would like to take advantage of them, I suggest you run composer update 😎  

Screenshot 2020-09-02 at 18 38 47

johanvanhelden:
Amazing! Thanks for the work @claudiodekker ! I really hope it gets merged!

claudiodekker:
@johanvanhelden For now, you can [require this package instead of using this fork](https://github.com/claudiodekker/inertia-laravel-testing).
I needed it in a production app and didn't want to rely on a fork of inertia, so decided to put it in a package 😅

johanvanhelden:
> @johanvanhelden For now, you can [require this package instead of using this fork](https://github.com/claudiodekker/inertia-laravel-testing).
> I needed it in a production app and didn't want to rely on a fork of inertia, so decided to put it in a package sweat_smile

Oh, sounds good, thanks for making a package!

innocenzi:
Good call making it a package, thanks @claudiodekker! I [added it](https://github.com/innocenzi/awesome-inertiajs/commit/6ce1c497f07227b50ca42ccc99f642b5f9083d93) to `awesome-inertia` until it gets deprecated.

johanvanhelden:
@claudiodekker since you seem to be working more closely with @reinink , do you know if there are any plans to make this part of the core now?

claudiodekker:
Hi @johanvanhelden, that's true, but for now we're fairly comfortable with keeping this [a separate package](https://github.com/claudiodekker/inertia-laravel-testing), which you can consider semi-official and expect to remain up to date with Inertia's changes.

That said, if and when we make this part of Inertia's core, I'll let you know by adding a message to this issue.

claudiodekker:
Moved to https://github.com/inertiajs/inertia-laravel/pull/220

PR #131, opened 2020-08-19
Fix #187 - Version MUST be a string, otherwise page refreshes

### LINKED ISSUE #187: Adding a variable to $page (like jetstream)
Jetstream adds an object to $page for easy accessing different variables, like team. How can I replicate this behaviour? I have looked through the code but I have not found where it's used.

claudiodekker:
See https://github.com/inertiajs/inertia/issues/187

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -46,7 +46,11 @@ class ResponseFactory
 
     public function getVersion()
     {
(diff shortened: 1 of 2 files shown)


reinink:
Awesome, thanks @claudiodekker! I appreciate the improved test coverage as well! 👌

PR #132, opened 2020-08-20
Add CI pipeline + fix multiple env issues

claudiodekker:
## Breaking Changes
This PR **drops Laravel 5.8 and earlier, as well as PHP 7.1 and earlier**. Below you'll find the reasons why, by going the issues I encountered and their solutions as to show why these changes were necessary.

- The first thing I did is to enable TravisCI and StyleCI on all supported PHP and Laravel versions, as to see what would fail. To do this, all that's required is to [run the tests against the 'lowest possible' dependency requirements, as well as the most up-to-date ones, on every single PHP version](https://github.com/inertiajs/inertia-laravel/blob/21ed43531912412747279ba9c378148704344e87/.travis.yml)

- The first issue that I ran into, is that [PHP 5.4](https://travis-ci.org/github/claudiodekker/inertia-laravel/jobs/731560911) and [PHP 5.5](https://travis-ci.org/github/claudiodekker/inertia-laravel/jobs/731560914) aren't even available on TravisCI anymore, due to already being deprecated a long time ago. **This means that we'd also have to drop support for PHP 5.4 and PHP 5.5**. Meaning that the lowest possible Laravel version would be 5.6

- Once these changes were made, the next issue I ran in to was regarding the 'new' [PHPUnit 8+ `setUp(): void` syntax requirement](https://phpunit.de/announcements/phpunit-8.html). Basically, this is PHPUnit forcing our hand in using PHPUnit 8.0, which requires PHP 7.1. **This lead me to also drop support for PHP 5.6 and 7.0**. In addition, this change bumps our minimum version of Laravel up to 5.8, [as that one supports both PHPUnit 7.5 and PHPUnit 8](https://github.com/laravel/framework/blob/5.8/composer.json#L88)

- Next, after running it again on PHP 7.1, [PHPUnit started complaining during runtime](https://travis-ci.org/github/claudiodekker/inertia-laravel/jobs/731568662#L396) that PHPUnit 8 is _really_ only supported with PHP 7.2+ (which was actually also [clearly mentioned on their website](https://phpunit.de/announcements/phpunit-8.html)), meaning we also **have to drop PHP 7.1 as well**. Luckily, this doesn't mean anything for our minimal Laravel version.

- Finally, I ran into an issue with Orchestra's Testbench on Laravel 5.8. Since Laravel 5.8 primarily uses PHPUnit 7.5 (even though it supports PHPUnit 8.0), Testbench went with the PHPUnit 7+ syntax, meaning [our PHPUnit 8+ tests will also fail on this version](https://travis-ci.org/github/claudiodekker/inertia-laravel/jobs/731568663#L333). Because of this Testbench decision / limitation, this also means we have to drop support for Laravel 5.8, making our minimum supported version Laravel 6.

## Features:
- Prevents the exporting of tests etc. into other people's applications ([using `.gitattributes` export-ignore](https://madewithlove.com/gitattributes/))
- `.editorconfig` for consistent codestyle
- StyleCI support (laravel codestyle)
- [Github Actions Testing support (Laravel 6+, PHP 7.2+)](https://github.com/claudiodekker/inertia-laravel/actions/runs/286946946)


claudiodekker:
This PR will cause a broken build, because testing against [PHP <= 7.1 will fail](https://travis-ci.org/github/claudiodekker/inertia-laravel-testing/jobs/723744981) for the following reasons:
- The test suite uses [the PHPUnit 8+ 'format'](https://phpunit.de/announcements/phpunit-8.html) (e.g. [`public function setUp(): void`](https://github.com/inertiajs/inertia-laravel/blob/master/tests/TestCase.php#L15)), whereas [PHP <= 7.1 doesn't support the `void` return type](https://wiki.php.net/rfc/void_return_type)
- The PHPUnit 8+ format is the 'only way' forward as of [Laravel 6](https://github.com/laravel/laravel/blob/6.x/composer.json) ([here's Laravel 5.8 for comparison](https://github.com/laravel/laravel/blob/5.8/composer.json))

---

In addition to the above, any PHP 7.2 tests **will not actually test against the most recent versions of Laravel**, because of the following reasons:
- The lowest possible 'required' version of PHPUnit is version 8, because:
- The lowest possible 'required' version of TestBench is [version 3.9](https://github.com/orchestral/testbench/blob/3.9/composer.json), [because of a 'limited' dependency In this repo](https://github.com/inertiajs/inertia-laravel/blob/master/composer.json#L31).
- The lowest possible version of Laravel is [version 6, because the of requirements in TestBench 3.9](https://github.com/orchestral/testbench/blob/3.9/composer.json)

Screenshot 2020-09-03 at 14 22 13

composer.json
```diff
@@ -25,10 +25,11 @@
         }
     },
     "require": {
-        "laravel/framework": "^5.4|^6.0|^7.0|^8.0"
```
reinink, line 28: Didn't you just add support for PHP 5.4 (the middleware)? 🤔

claudiodekker: Yeah, I did.. 😢

claudiodekker:
Moved testing from Travis CI to Github Actions, which is significantly faster (multiple minutes) 🚀 
(Based on https://github.com/laravel/jetstream/blob/1.x/.github/workflows/tests.yml)

claudiodekker:
Fixed in #220

PR #135, opened 2020-09-01
Automatically share validation errors

reinink:
This updates this adapter to automatically share validation errors with Inertia.js, instead of making Laravel users do this manually, like this:

```php
Inertia::share([
    'errors' => function () {
        return Session::get('errors')
            ? Session::get('errors')->getBag('default')->getMessages()
            : (object) [];
    },
]);
```

The key difference with the above method is that Inertia will automatically return only *one* error per field, instead of the Laravel default, which is an array of errors.

```js
{
  name: 'The name field is required.',
  email: 'Not a valid email address.',
}
```

This is simpler, and honestly, I've literally never displayed multiple errors per field. This seems to be the consensus [on Twitter](https://twitter.com/reinink/status/1300516500105842689) as well.

That said, if folks prefer the other behaviour (multiple errors per field), all they need to do is manually share the validation errors like before (using the snippet above). This package's service provider will only share the validation errors if the `errors` key has not already been used. This behaviour also makes this a non-breaking change.

The goal of this change is to make Inertia.js easier to setup in Laravel applications, and since sharing validation errors is so common, this seems worthwhile to bake into the adapter.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -6,6 +6,7 @@ use Illuminate\Http\Request;
 use Illuminate\Routing\Router;
 use Illuminate\Contracts\Http\Kernel;
 use Illuminate\Support\Facades\Blade;
+use Illuminate\Support\Facades\Session;
 use Illuminate\Support\ServiceProvider as BaseServiceProvider;
 
 class ServiceProvider extends BaseServiceProvider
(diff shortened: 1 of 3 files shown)


herpaderpaldent:
thank you @reinink for this PR, i really like to see this change merged. May i point out that the documentation should be updated accordingly?

PR #137, opened 2020-09-01
Add support for (multiple) error bags

claudiodekker:
## Added
- Support for multiple error bags
- Prevent passing unexpected values.
- Extra testing for 'default' (no-errors) scenario
- [**EXTRA / SUGGESTED**: `Response::withErrors(..)`.](https://github.com/claudiodekker/inertia-laravel/pull/1/files#diff-9cd62f6ac47a5969c6156a947d50b7e4R191)

## Fixed
- Use `ViewErrorBag` instead of `MessageBag` (it works [due to proxy-calls](https://github.com/laravel/framework/blob/17777a92da9b3cf0026f26462d289d596420e6d0/src/Illuminate/Support/ViewErrorBag.php#L93-L96), but [it's not really intended to be used that way](https://github.com/laravel/framework/blob/7.x/src/Illuminate/View/View.php#L212-L230))

## Remaining / newly found issues
- When an actual `ValidationException` gets thrown, error messages **won't** get formatted correctly due to the way the ValidationException is defined/handled by the ExceptionHandler. [Here's a failing test that demonstrates the issue](https://github.com/inertiajs/inertia-laravel/compare/validation-errors...claudiodekker:validation-errors-failing-validator).

There's a few solutions to this that I can tell, and neither of them aren't really pretty:
- Extend the Application's ExceptionHandler, then swap it out in the container. This will allow us to intercept the ValidationException and re-format the messages.

- Swap out the Validator for our own, so we can throw a custom ValidationException.

- Tell the user to add a call to their Application's ExceptionHandler, so we don't have to override/extend any core-logic (less maintenance, more work for the end-user)

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -4,6 +4,8 @@ namespace Inertia;
 
 use Illuminate\Http\Request;
 use Illuminate\Routing\Router;
+use Illuminate\Support\MessageBag;
+use Illuminate\Support\ViewErrorBag;
 use Illuminate\Contracts\Http\Kernel;
 use Illuminate\Support\Facades\Blade;
 use Illuminate\Support\Facades\Session;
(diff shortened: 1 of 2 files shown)


reinink:
Thanks so much for your help here @claudiodekker! I've simplified things a little, but the concept was right. 👌

PR #139, opened 2020-09-08
Update ServiceProvider@registerBladeDirective

halilcakar:
Enables to change parent div id's, defaulted to `app`


reinink:
Reasonable addition, thanks! 👍

Issue #140, opened 2020-09-09
@inertia directive broken


robsontenorio:
Probably related to #139  cc @halilcakar 

If no param passed it breaks

![image](https://user-images.githubusercontent.com/118955/92546810-d5a28400-f229-11ea-9619-d7e38cfa25be.png)

![image](https://user-images.githubusercontent.com/118955/92546795-ce7b7600-f229-11ea-8d4d-c47c43d5e60c.png)

Params seems to be mandatory 

![image](https://user-images.githubusercontent.com/118955/92546882-02569b80-f22a-11ea-8b1b-567fb2f2c299.png)

![image](https://user-images.githubusercontent.com/118955/92546919-19958900-f22a-11ea-9c59-1213cd5321de.png)

reinink:
Ugh, never merge something in without testing. 🤦‍♂️

Let me fix this.

robsontenorio:
@reinink Oww! faster than MacDonalds ! Thanks :)

halilcakar:
Woah, Sorry for that, I've also didn't have the chance the test thanks @robsontenorio

reinink:
I'm not sure sure how exactly you handle a plain string argument. I think I'm going to revert for now. Feel free to resubmit a working version.

reinink:
Reverted for now (https://github.com/inertiajs/inertia-laravel/releases/tag/v0.2.11).

reinink:
Thanks for reporting @robsontenorio! 👍

halilcakar:
Aight I've just tested out and it's working like following:

```php
Blade::directive('inertia', function ($id) {
    if(func_get_arg(0) === "") {
        $id = 'app';
    }

    return '<div id="' . $id . '" data-page="{{ json_encode($page) }}"></div>';
});
```

I'm not sure if you would like this type of checks @reinink, but for now the best i can think of.

PR #141, opened 2020-09-09
Re-implement & test #139

claudiodekker:
Fixes #139, #140 

This `compileBlade` function is mostly [based off of what Laravel uses in their own test suite](https://github.com/laravel/framework/blob/7.x/tests/View/Blade/AbstractBladeTestCase.php). 

I went with this, as rendering a view with conditions for both test-cases became too messy (and blade view caching kicks in as well, giving inconsistent test results):


claudiodekker:
Closed in favour of #186

PR #143, opened 2020-09-09
Add simple XHR response

lepikhinb:
This tiny PR adds an ability to re-use the same controllers for simple XHR requests as well as for Inertia ones.
Not sure if that's really a missing feature, but I just faced a case when I had to duplicate multiple controllers just to allow fetching data using axios without extra headers on input and extra data on output.

Example usage:
– Search
– Dynamic selects (v-select, select2)

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -84,6 +84,10 @@ class Response implements Responsable
             ]);
         }
 
+        if ($request->isXmlHttpRequest()) {
+            return new JsonResponse($props);
+        }
+
         return ResponseFactory::view($this->rootView, $this->viewData + ['page' => $page]);
     }
 }
--- a/tests/ResponseTest.php
+++ b/tests/ResponseTest.php
@@ -53,6 +53,20 @@ class ResponseTest extends TestCase
         $this->assertSame('123', $page->version);
     }
 
+    public function test_xhr_non_inertia_response()
+    {
+        $request = Request::create('/user/123', 'GET');
+        $request->headers->add(['X-Requested-With' => 'XMLHttpRequest']);
+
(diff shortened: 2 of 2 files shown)


Wiejeben:
This is already possible by providing a `X-Inertia: true` header (see: https://inertiajs.com/the-protocol#inertia-responses), but I never really liked that.

reinink:
@lepikhinb This is an interesting idea, and I love the simplicity of the implementation. I am curious though, why not just make the request with Inertia using partial reloads to only get the data you need?

@Wiejeben I'm not really sure what you mean...

Wiejeben:
Nvm, I always assumed `X-Inertia: true` would make the Inertia return JSON instead of HTML. My bad!

reinink:
Thinking about this further...

While I think this is a neat idea, I'm worried about adding support for XHR in the Inertia responses. While this is (obviously) a simple little change, I'm worried what things it might open up (in terms of feature requests). We just had a chat with the "internals" team, and the feeling right now is this is probably best left out, especially since this is quite easy to handle manually in your controllers.

```php
{
    $props = [
        // ...
    ];

    if ($request->isXmlHttpRequest()) {
        return $props;
    }

    return Inertia::render('component', $props);
}
```

Or alternatively:

```php
{
    $props = [
        // ...
    ];

    return $request->inertia()
        ? Inertia::render('component', $props)
        : $props;   
}
```

Given that the need for this is somewhat edge-case, and given that there is a reasonable workaround for this, I'm going to say we leave this out for now.

Thanks either way for the contribution, your interest in Inertia.js, and for the fact you included tests with your PR! 💜

PR #145, opened 2020-09-12
Add Inertia::forceRedirect

### LINKED ISSUE #57: Add a method to force a redirect to a non-Inertia URL
Sometimes we want to redirect the user to a non-Inertia environment after a visit. 

Inertia already supports something similar under the hood, for asset versioning.

https://github.com/inertiajs/inertia/blob/da004ee147cf53a7f02804f50dc0a324ecb9fdec/src/inertia.js#L83-L85

We added this little helper function in our app, but it feels hacky:

```php
function redirectWithoutInertia(string $url)
{
    return response('', SymfonyResponse::HTTP_CONFLICT)->header('x-inertia-location', $url);
}
```

Would it make sense to add an `Inertia::forceRedirect($url)` method to inertia-laravel?

KABBOUCHI:
Example usages:
- When you want to visit a URL outside Laravel app
- When you want to visit a URL inside Laravel app, but its not a inertia page 

Closes #57

### DIFF
--- a/src/Inertia.php
+++ b/src/Inertia.php
@@ -11,6 +11,7 @@ use Illuminate\Support\Facades\Facade;
  * @method static void version($version)
  * @method static int|string getVersion()
  * @method static \Inertia\Response render($component, $props = [])
+ * @method static \Illuminate\Http\Response|\Illuminate\Http\RedirectResponse forceRedirect($to, $status = 302, $headers = [], $secure = null)
  *
  * @see \Inertia\ResponseFactory
  */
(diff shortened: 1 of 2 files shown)


reinink:
Hey thanks for this @KABBOUCHI! However, I've decided to go with a slightly different implementation.

My biggest concern with this implementation is the fallback to a regular redirect. At first, I thought that was a good idea too, but I actually think this is too opinionated. I don't think that we can assume that just because it's not an Inertia visit, that a redirect is the right answer. Maybe the right answer is a plain JSON response. I think it's better to leave this decision up to the developer.

I appreciate you trying to fix this! ❤️

I'll get my implementation merged in shortly. 👍

PR #150, opened 2020-09-28
Automatically handle CSRF Errors in Laravel 8+

claudiodekker:
We _could_ also handle other Exceptions and automatically render to the 'Error' component, but I figured I'd leave that up to you. The nice thing, is that this is rendered by Laravel's core, meaning users could still override it easily by handling it in their application's `ExceptionHandler`, similar to how it's being done today.

---

```php
$statusCode = $exception->getStatusCode();
$previous = $exception->getPrevious();

if ($statusCode === 419 && $previous instanceof TokenMismatchException) {
    return Redirect::back()->with([
        'error' => Config::get('inertia.csrf_error', 'The page expired, please try again.')
    ]);
}

if (Config::get('inertia.error.enabled', false)
    && app()->environment(Config::get('inertia.error.environments', 'production'))
    && in_array($statusCode, Config::get('inertia.error.status_codes', [500, 503, 404, 403]))
) {
    return Inertia::render(Config::get('inertia.error.component', 'Error'), [
        'status' => $statusCode, 'exception' => $previous
    ])
    ->toResponse(request())
    ->setStatusCode($statusCode);
}
```

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -9,13 +9,19 @@ use Illuminate\Contracts\Http\Kernel;
 use Illuminate\Support\Facades\Blade;
 use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\Session;
+use Illuminate\Support\Facades\Redirect;
+use Illuminate\Session\TokenMismatchException;
+use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Support\ServiceProvider as BaseServiceProvider;
+use Symfony\Component\HttpKernel\Exception\HttpExceptionInterface;
 
 class ServiceProvider extends BaseServiceProvider
 {
     public function register()
     {
         $this->app->singleton(ResponseFactory::class);
+
+        $this->registerExceptionHandlers();
     }
 
     public function boot()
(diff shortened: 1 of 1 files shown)


reinink:
Hmm, as discussed on Discord, I don't think I really want this baked in...at least not yet. This feels too opinionated to me. Even the `error` shared key isn't required here...and someone can totally choose to do something else. In all likelihood this would result in folks being confused why their Inertia visits are going back to the same page. Further, why would we add automatic handling of this, and not the other error responses?

I think, for now, I'd rather just provide a good error handling implementation in the docs.

src/ServiceProvider.php
```diff
@@ -78,4 +84,24 @@ protected function shareValidationErrors()
...
+        // In earlier version of Laravel, you'll have to register
+        // these yourself in your app's exception handler.
+        if (! method_exists($handler, 'renderable')) {
+            return;
+        }
+
+        $handler->renderable(function (HttpExceptionInterface $exception) {
+            if ($exception->getStatusCode() === 419 && $exception->getPrevious() instanceof TokenMismatchException) {
+                return Redirect::back()->with([
+                    'error' => Config::get('inertia.csrf_error', 'The page expired, please try again.'),
```
RobertBoes, line 102:
```suggestion
                    'error' => Config::get('inertia.csrf_error', __('The page expired, please try again.')),
```

I think it would be good to also translate this text by default

PR #154, opened 2020-10-02
Add new Inertia::location($url) method

### LINKED ISSUE #57: Add a method to force a redirect to a non-Inertia URL

reinink:
Closes #57.

This PR adds a new `Inertia::location($url)` method, which returns a "409 Conflict" response with the `X-Inertia-Location` header set. This will cause a full page visit (`location.href = url`) to happen client-side by Inertia. This works since this behaviour is already in place for automatic asset refreshing.

```php
return Inertia::location('https://www.google.com');
```

I really struggled with naming this method, but in the end decided that `location()` made the most sense, because:

1. There is no confusion here with this being a redirect. I was worried that adding a method with the word "redirect" it in might cause developers to use this method instead of just regular redirects (even when a regular redirect is totally the right approach).
2. It doesn’t feel like you’re doing something wrong. We considered `hardVisit($url)` and `abortTo($url)`, but that almost implied that you were doing something incorrect.
3. It’s short.
4. It aligns really well with what's actually happening (a `window.location = url`) visit.
5. It even matches the header we are already using for this (`X-Inertia-Location`).

Credit to @adamwathan for coming up with this method name. 🥰

### DIFF
--- a/src/Inertia.php
+++ b/src/Inertia.php
@@ -11,6 +11,7 @@ use Illuminate\Support\Facades\Facade;
  * @method static void version($version)
  * @method static int|string getVersion()
  * @method static \Inertia\Response render($component, $props = [])
+ * @method static \Illuminate\Http\Response location($url)
  *
  * @see \Inertia\ResponseFactory
  */
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -5,6 +5,7 @@ namespace Inertia;
 use Closure;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\App;
+use Illuminate\Support\Facades\Response as BaseResponse;
 use Illuminate\Support\Traits\Macroable;
 use Illuminate\Contracts\Support\Arrayable;
 
@@ -65,5 +66,10 @@ class ResponseFactory
             $this->rootView,
             $this->getVersion()
         );
+    }
+
+    public function location($url)
+    {
+        return BaseResponse::make('', 409, ['X-Inertia-Location' => $url]);
     }
 }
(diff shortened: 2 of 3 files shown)


reinink:
Pinging @bknoles, @ledermann, @BrandonShar. Gents, think we could get the same method added to the Rails adapter? 🙏

BrandonShar:
Cool stuff! I'll need to think a bit about the right syntax, but I think I can get this in today.

reinink:
@BrandonShar Awesome! Someone already suggested the implementation [here](https://github.com/inertiajs/inertia-laravel/issues/57#issuecomment-570147828). Just need to update to a better name than "redirect_without_inertia". It would be great to use "location" just like in the Laravel adapter. Something like `inertia.location('http://google.ca/')`. (Fair warning, I don't know Ruby at all lol.)

BrandonShar:
Yeah, that solution looks pretty solid. I was wondering if this should be a part of the renderer, but I don't think so. Just making sure it has the right "rails-y" feel since we don't use the `Inertia` object directly like the laravel version does.

Does `inertia_location('http://google.ca/')` seem right to you? That would line up with our existing `inertia_share` helper.

BrandonShar:
(and rail's existing `redirect_to('http://google.ca')` pattern)

reinink:
Yes, perfect, I was wondering if you would suggest `inertia_location()`. 👍

PR #155, opened 2020-10-03
Preserve url hash on navigation

rgehan:
This PR is an attempt at fixing https://github.com/inertiajs/inertia/issues/85.

It goes with its inertiajs/inertia counterpart: https://github.com/inertiajs/inertia/pull/244

It simply appends the content of the `X-Inertia-Hash` header to the URL that is returned to the frontend.

This ensures that navigating to `/foo/bar#some-hash` results in the URL being `/foo/bar#some-hash` and not just `/foo/bar`.

---

If you are happy with the proposed changes, here are a few things that will need to be done too:
- [ ] Support the same behavior in inertiajs/inertia-rails
- [ ] Update PingCRM to demonstrate this behavior

---

Feel free to suggest any improvement :)

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -70,10 +70,12 @@ class Response implements Responsable
             }
         });
 
+        $hash = $request->header('X-Inertia-Hash', '');
+
         $page = [
             'component' => $this->component,
             'props' => $props,
-            'url' => $request->getRequestUri(),
+            'url' => $request->getRequestUri() . $hash,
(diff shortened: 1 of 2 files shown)


reinink:
Closing this in favour of https://github.com/inertiajs/inertia/pull/257 (which is almost done).

Thanks so much for bringing this to my attention, and pushing me to fix this @rgehan!

PR #158, opened 2020-10-07
Adds support for adding middleware to multiple middleware groups

ksassnowski:
In one of my projects I needed to add a custom middleware group--in addition to `web`--that also needed to work with Inertia. Currently I have to manually add Inertia's middleware to my custom middleware stack. 

This PR adds support for specifying a list of middleware groups that the middleware should be pushed to. The default behavior did not change and it is still possible to assign a single value to `inertia.middleware_group`.

Getting the test to run was a bit finicky because I have to overwrite the config value. By the time the test runs, however, the ServiceProvider has already been registered and changing the config inside the test won't have an effect anymore. The only way I could get the test to run was to manually instantiate the service provider again and run the `boot` method. If anyone knows a better way to do this, please let me know.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace Inertia;
 
+use Illuminate\Support\Arr;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Router;
 use Illuminate\Support\Collection;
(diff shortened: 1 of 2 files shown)


reinink:
Hey thanks for the contribution! I don't think this change is really necessary though. If your app requires the Inertia middleware to be registered in multiple middleware groups, you can simply manually register it yourself. That feels like the easier approach than to make this auto registration more complicated.

Maybe this is more of a documentation issue. It would probably be good to update the docs to explain this.

ksassnowski:
Fair enough. We stumbled over this because the middleware was moved from the global middleware stack to the web group a few release ago.

Updating the docs is probably a good idea. I'll see if I can submit a PR there then.

ksassnowski:
Speaking of, since this seems like a Laravel specific thing (might affect Rails too, idk), where inside the documentation should this go?

reinink:
Great question. I wonder if we add a new "Middleware" section right before the "Root template" which explains that both the Laravel and Rails adapters automatically register the Inertia middleware, but then show a code sample of how to do this manually.

reinink:
Oh, and I'm talking about this page: https://inertiajs.com/server-side-setup

PR #161, opened 2020-10-09
Inertia Middleware: A better home for user-provided defaults.

claudiodekker:
This PR introduces a new way of setting up and sharing application defaults with Inertia. 

Up until this point, [we've always used (and recommended) the `AppServiceProvider`](https://github.com/inertiajs/pingcrm/blob/master/app/Providers/AppServiceProvider.php#L30-L59) as the place to do this, but doing so has a few negative impacts, most important one being that Inertia gets bootstrapped even when it does not have to (e.g. when running inside the Console).

By using Middleware for this instead, we not only solve this problem, but it becomes possible to provide different defaults for each route (group), as you're free to (re)configure these middleware as you see fit!

To use:
- Run `php artisan inertia:middleware` to generate a new `HandleInertiaRequests` middleware.
- Add the `\App\Http\Middleware\HandleInertiaRequests::class` to your application's Middleware stack.

Example implementation: https://github.com/inertiajs/pingcrm/pull/88/files


reinink:
Thanks so much for working on this @claudiodekker!

So, I don't really understand the need for the trait here. I think it would be nice to just use regular inheritance instead. I've hacked on this a little tonight, and got it working the way I was hoping Please find both a proposed new Inertia adapter middleware below, and an example implementation for Ping CRM.

## Adapter: `Inertia/Middleware`

```php
<?php

namespace Inertia;

use Closure;
use Illuminate\Http\Request;
use Illuminate\Support\Collection;
use Inertia\Inertia;
use Symfony\Component\HttpFoundation\RedirectResponse;
use Symfony\Component\HttpFoundation\Response;

class Middleware
{
    public function handle($request, Closure $next)
    {
        Inertia::version($this->version($request));
        Inertia::share($this->share($request));

        $response = $next($request);
        $response = $this->checkVersion($request, $response);
        $response = $this->changeRedirectCode($request, $response);

        return $response;
    }

    public function checkVersion(Request $request, Response $response)
    {
        if ($request->header('X-Inertia') &&
            $request->method() === 'GET' &&
            $request->header('X-Inertia-Version', '') !== Inertia::getVersion()
        ) {
            if ($request->hasSession()) {
                $request->session()->reflash();
            }

            return new Response('', 409, ['X-Inertia-Location' => $request->fullUrl()]);
        }

        return $response;
    }

    public function changeRedirectCode(Request $request, Response $response)
    {
        if ($request->header('X-Inertia') &&
            $response instanceof RedirectResponse &&
            $response->getStatusCode() === 302 &&
            in_array($request->method(), ['PUT', 'PATCH', 'DELETE'])
        ) {
            $response->setStatusCode(303);
        }

        return $response;
    }

    public function version(Request $request)
    {
        return null;
    }

    public function share(Request $request)
    {
        return [
            'errors' => function () use ($request) {
                return $this->resolveValidationErrors($request);
            }
        ];
    }

    public function resolveValidationErrors(Request $request)
    {
        if (! $request->session()->has('errors')) {
            return (object) [];
        }

        return (object) Collection::make($request->session()->get('errors')->getBags())->map(function ($bag) {
            return (object) Collection::make($bag->messages())->map(function ($errors) {
                return $errors[0];
            })->toArray();
        })->pipe(function ($bags) {
            return $bags->has('default') ? $bags->get('default') : $bags->toArray();
        });
    }
}
```

## User-land: `App\Http\Middleware\Inertia`

```php
<?php

use Closure;
use Illuminate\Http\Request;
use Inertia\Middleware;

class Inertia extends Middleware
{
    public function version(Request $request)
    {
        return md5_file(public_path('mix-manifest.json'));
    }
    
    public function share(Request $request)
    {
        return array_merge(parent::share($request), [
            'auth' => function () use ($request) {
                return [
                    'user' => $request->user() ? [
                        'id' => $request->user()->id,
                        'first_name' => $request->user()->first_name,
                        'last_name' => $request->user()->last_name,
                        'email' => $request->user()->email,
                        'role' => $request->user()->role,
                        'account' => [
                            'id' => $request->user()->account->id,
                            'name' => $request->user()->account->name,
                        ],
                    ] : null,
                ];
            },
            'flash' => function () use ($request) {
                return [
                    'success' => $request->session()->get('success'),
                    'error' => $request->session()->get('error'),
                ];
            },
        ]);
    }
}
```

claudiodekker:
Hey @reinink 👋 ,

Yeah, that was actually the first thing I tried, but for some reason it wouldn't work? I always got into the situation where it would call `$this->share($request)` on the 'base' Middleware, and wouldn't pick up on the overloaded/extended instance. (And I'm sure I created an instance of the 'extended' class, and not the base class..)

Anyway, I probably missed something, because I tried using what you suggested this morning, and it worked just fine 🎉  
I've updated the PR to reflect those changes, as well as some other things. Pretty satisfied with the result!

src/Middleware.php
```diff
@@ -3,31 +3,127 @@
...
+    /**
+     * Determines the current Inertia asset version hash.
+     *
+     * @see https://inertiajs.com/asset-versioning
+     * @param  \Illuminate\Http\Request  $request
+     * @return string|null
+     */
+    public function version(Request $request)
     {
+        //
```
claudiodekker, line 22: Also, I was thinking.. what if we would do this in the _base_ Middleware?:

```suggestion
        // When we are running on Laravel Vapor, asset URLs are automatically updated
        // when the application is being deployed. Because of this, we do not need
        // to use any files for hashing, as we can simply use this URL instead.
        if (config('app.asset_url')) {
            return md5(config('app.asset_url'));
        }

        // Alternatively, when we are running in a regular (non-serverless) environment
        // we'll attempt to use the Laravel Mix asset manifest to generate our hash.
        if (file_exists($manifest = public_path('mix-manifest.json'))) {
            return md5_file($manifest);
        }
```

Since this adapter is Laravel specific to begin with, this would allow users to enable it using an one-liner, and would also skip the need to add this directly into Jetstream.

In addition, it still won't be enabled by default, because the `inertia:middleware` command generates the Middleware without the `parent::version($request)` call

src/Middleware.php
```diff
@@ -3,31 +3,127 @@
...
 use Closure;
-use Illuminate\Support\Facades\Response;
-use Symfony\Component\HttpFoundation\RedirectResponse as Redirect;
+use Illuminate\Http\Request;
+use Illuminate\Support\Collection;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Response;
 
-class Middleware
+abstract class Middleware
```
reinink, line 11: Does it need to be abstract? Like technically if someone wanted to use it straight up, they could.

src/Middleware.php
```diff
@@ -3,31 +3,127 @@
...
 {
-    public function handle($request, Closure $next)
+    /**
+     * Determines the current Inertia asset version hash.
```
reinink, line 14: I would just say:

```
Determines the current asset version.
```

Since *technically* it doesn't have to be a hash (even if we end up having a default implementation that does do this).

src/Middleware.php
claudiodekker: Yeah, that's true. 

My thinking here was that it would provide a single way to do it, but maybe that's indeed a bit too limiting because we're not doing anything special here.

src/Middleware.php
reinink: Honestly, I'm tempted to just do this, and have it enabled by default (not commented out in the generated middleware), and folks can always change it if they need to.

This seems like a totally reasonable way of determining the asset version in an automated way.

stubs/Middleware.blade.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace App\Http\Middleware;
+
+use Illuminate\Http\Request;
+use Inertia\Middleware;
+
+class {{ $name }} extends Middleware
+{
+    /**
+     * Determines the current Inertia asset version hash.
```
reinink, line 11: Tweak here as well:

> Determines the current asset version.

src/Middleware.php
```diff
@@ -3,31 +3,127 @@
...
+     * @param  \Illuminate\Http\Request  $request
+     * @return string|null
+     */
+    public function version(Request $request)
     {
+        //
+    }
+
+    /**
+     * Defines the Inertia props that are shared by default.
```
reinink, line 26: I don't think we need the word "Inertia" here:

> Defines the props that are shared by default.

src/Middleware.php
```diff
@@ -3,31 +3,127 @@
...
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  Closure  $next
+     * @return Response
+     */
+    public function handle(Request $request, Closure $next)
+    {
+        Inertia::share($this->share($request));
+
+        Inertia::version(function () use ($request) {
```
reinink, line 63: Does this need to be in a closure, since this now only runs when the middleware runs?

claudiodekker: It doesn't, but it allows the user to not have to care about it in their `version()` function, and still have it lazy evaluate out of the box 🎉

reinink: Right, but even for end users, if this is done in a middleware, why would they need it to be lazily evaluated if it's going to be immediately resolved?

claudiodekker: It won't, because the controller action still has to happen first, meaning that there might still be a case where a redirect happens instead of an Inertia response being rendered. 

In that scenario, resolving it immediately would mean unnecessarily loading a file & hashing it.

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
...
+        $response = $this->checkVersion($request, $response);
+        $response = $this->changeRedirectCode($request, $response);
+
+        return $response;
+    }
 
-        if ($request->method() === 'GET' && $request->header('X-Inertia-Version', '') !== Inertia::getVersion()) {
+    /**
+     * In the event that the asset version changes, Inertia will automatically reload
+     * the page in order to ensure the application keeps working for the end user.
```
reinink, line 76:
```suggestion
     * In the event that the assets change, initiate a
     * client-side location visit to force an update.
```

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
...
+        if ($request->header('X-Inertia') &&
+            $request->method() === 'GET' &&
+            $request->header('X-Inertia-Version', '') !== Inertia::getVersion()
+        ) {
             if ($request->hasSession()) {
                 $request->session()->reflash();
             }
 
-            return Response::make('', 409, ['X-Inertia-Location' => $request->fullUrl()]);
+            return new Response('', 409, ['X-Inertia-Location' => $request->fullUrl()]);
```
reinink, line 92: So obvious, but we can now do this:

```suggestion
            return Inertia::location($request->fullUrl());
```

src/Middleware.php
reinink: Yup, okay, fair enough! 👍

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
...
+        ) {
             $response->setStatusCode(303);
         }
 
         return $response;
     }
+
+    /**
+     * Resolves and prepares validation errors in such a way
+     * that they are easier to use in the Inertia client.
```
reinink, line 121:
```suggestion
     * Resolves and prepares validation errors in such
     * a way that they are easier to use client-side.
```

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
...
+            return new Response('', 409, ['X-Inertia-Location' => $request->fullUrl()]);
         }
 
-        if ($response instanceof Redirect && $response->getStatusCode() === 302 && in_array($request->method(), ['PUT', 'PATCH', 'DELETE'])) {
+        return $response;
+    }
+
+    /**
+     * Changes the status code during an Inertia redirect, so that the browser
+     * makes the request as GET instead, avoiding unsupported method calls.
```
reinink, line 100:
```suggestion
     * Changes the status code during redirects, ensuring they are made as
     * GET requests, preventing "MethodNotAllowedHttpException" errors.
```

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
...
+     * @return object
+     */
+    public function resolveValidationErrors(Request $request)
+    {
+        if (! $request->session()->has('errors')) {
+            return (object) [];
+        }
+
+        return (object) Collection::make($request->session()->get('errors')->getBags())->map(function ($bag) {
+            return (object) Collection::make($bag->messages())->map(function ($errors) {
```
reinink, line 133: We're already using both the `public_path()` and `config()` helpers, might as well use the `collect()` helper as well. 😆

```suggestion
        return (object) collect($request->session()->get('errors')->getBags())->map(function ($bag) {
            return (object) collect($bag->messages())->map(function ($errors) {
```

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
 namespace Inertia;
 
 use Closure;
-use Illuminate\Support\Facades\Response;
-use Symfony\Component\HttpFoundation\RedirectResponse as Redirect;
+use Illuminate\Http\Request;
+use Illuminate\Support\Collection;
```
reinink, line 7:
```suggestion
```

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
...
+     * @param  Closure  $next
+     * @return Response
+     */
+    public function handle(Request $request, Closure $next)
+    {
+        Inertia::share($this->share($request));
+
+        Inertia::version(function () use ($request) {
+            return $this->version($request);
+        });
```
reinink, line 65: We've been putting "version" before "share" everywhere, so why not be consistent here as well?
```suggestion
        Inertia::version(function () use ($request) {
            return $this->version($request);
        });
        
        Inertia::share($this->share($request));
```

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
...
+     *
+     * @param  Request  $request
+     * @param  Response  $response
+     * @return Response
+     */
+    public function changeRedirectCode(Request $request, Response $response)
+    {
+        if ($response instanceof RedirectResponse &&
+            $request->header('X-Inertia') &&
+            $response->getStatusCode() === 302 &&
```
reinink, line 110: Why are we event checking for the `RedirectResponse` class here? All we really care about is if the status code is `302`, which we already check.

```suggestion
        if ($request->header('X-Inertia') &&
            $response->getStatusCode() === 302 &&
```

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
 namespace Inertia;
 
 use Closure;
-use Illuminate\Support\Facades\Response;
-use Symfony\Component\HttpFoundation\RedirectResponse as Redirect;
+use Illuminate\Http\Request;
+use Illuminate\Support\Collection;
+use Symfony\Component\HttpFoundation\RedirectResponse;
```
reinink, line 8: No longer needed (as per other comment).

```suggestion
```

src/Middleware.php
claudiodekker: That's.. a very good point.

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
...
+     * @see https://inertiajs.com/asset-versioning
+     * @param  \Illuminate\Http\Request  $request
+     * @return string|null
+     */
+    public function version(Request $request)
     {
-        $response = $next($request);
+        // When we are running on Laravel Vapor, asset URLs are automatically updated
+        // when the application is being deployed. Because of this, we do not need
+        // to use any files for hashing, as we can simply use this URL instead.
```
reinink, line 24: I don't think we actually need this explanation. It's fairly self explanatory. 🤷‍♂️ 
```suggestion
```

src/Middleware.php
```diff
@@ -3,31 +3,138 @@
...
+        if (config('app.asset_url')) {
+            return md5(config('app.asset_url'));
+        }
 
-        if (!$request->header('X-Inertia')) {
-            return $response;
+        // Alternatively, when we are running in a regular (non-serverless) environment
+        // we'll attempt to use the Laravel Mix asset manifest to generate our hash.
```
reinink, line 30: Not needed IMHO.
```suggestion
```

tests/middleware/ExampleMiddleware.php
```diff
@@ -0,0 +1,48 @@
...
+    protected $shared = [];
+
+    public function __construct($version = null, $shared = [])
+    {
+        $this->version = $version;
+        $this->shared = $shared;
+    }
+
+    /**
+     * Determines the current Inertia asset version hash.
```
reinink, line 26:
```suggestion
     *  Determines the current asset version.
```

tests/middleware/ExampleMiddleware.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param  \Illuminate\Http\Request  $request
+     * @return string|null
+     */
+    public function version($request)
+    {
+        return $this->version;
+    }
+
+    /**
+     * Defines the Inertia props that are shared by default.
```
reinink, line 38:
```suggestion
     * Defines the props that are shared by default.
```

src/Middleware.php
claudiodekker: Not sure, I personally feel like it might seem like an 'odd' check, as the asset URL might seem fairly generic otherwise?

reinink: I don't think so, the `asset_url` is a "Laravel thing": https://github.com/laravel/laravel/blob/6bfe68365d0ab1301dfae87867814048440d6a78/config/app.php#L57

And, because this could be in non-Vapor situations as well, it seems odd to have a Vapor specific comment here.

I think we're good to go without the explanation. 👍

src/Middleware.php
claudiodekker: How's this?:

```suggestion
     * Resolves and prepares validation errors in such a way
     * that they become easier to use on the client-side.
```

tests/middleware/ExampleMiddleware.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Inertia\Tests\middleware;
```
reinink, line 3:
```suggestion
namespace Inertia\Tests\Middleware;
```

tests/ControllerTest.php
```diff
@@ -2,39 +2,33 @@
 
 namespace Inertia\Tests;
 
-use Inertia\Response;
+use Illuminate\Session\Middleware\StartSession;
+use Illuminate\Support\Facades\Route;
 use Inertia\Controller;
-use Illuminate\Http\Request;
-use Illuminate\Routing\Route;
+use Inertia\Tests\middleware\ExampleMiddleware;
```
reinink, line 8:
```suggestion
use Inertia\Tests\Middleware\ExampleMiddleware;
```

tests/MiddlewareTest.php
```diff
@@ -2,89 +2,149 @@
...
-use Illuminate\Foundation\Testing\TestResponse;
 use Illuminate\Http\Request;
+use Illuminate\Session\Middleware\StartSession;
+use Illuminate\Support\Facades\Route;
+use Illuminate\Support\Facades\Session;
+use Illuminate\Support\MessageBag;
+use Illuminate\Support\ViewErrorBag;
 use Inertia\Inertia;
-use Inertia\Middleware;
+use Inertia\Tests\middleware\ExampleMiddleware;
```
reinink, line 12:
```suggestion
use Inertia\Tests\Middleware\ExampleMiddleware;
```

tests/ResponseFactoryTest.php
```diff
@@ -2,8 +2,12 @@
 
 namespace Inertia\Tests;
 
-use Inertia\ResponseFactory;
 use Illuminate\Http\Response;
+use Illuminate\Session\Middleware\StartSession;
+use Illuminate\Support\Facades\Route;
+use Inertia\Inertia;
+use Inertia\ResponseFactory;
+use Inertia\Tests\middleware\ExampleMiddleware;
```
reinink, line 10:
```suggestion
use Inertia\Tests\Middleware\ExampleMiddleware;
```

PR #167, opened 2020-10-15
Ability to set root view on the new middleware

Cannonb4ll:


src/Middleware.php
```diff
@@ -57,6 +59,8 @@ public function handle(Request $request, Closure $next)
 
         Inertia::share($this->share($request));
 
+        Inertia::setRootView($this->rootView($request) ?? $this->rootView);
```
claudiodekker, line 62: @reinink I just checked, but this would be a breaking change for PHP <= 7. 

I already made a PR earlier ([testing & CI pipeline](https://github.com/inertiajs/inertia-laravel/pull/132)) in which I basically came to the conclusion that if we want full test coverage, we can't really support PHP 7.1 and earlier anyway, so I don't think this is really a big issue. Thoughts?

Cannonb4ll: I'd say drop anything lower than that. Its actually one of Spatie's standards to drop older versions rather quickly.

claudiodekker: @Cannonb4ll Hm, actually, I think we can avoid this breaking change entirely by returning `$this->rootView` from the `rootView()` method by default:
```suggestion
        Inertia::setRootView($this->rootView($request));
```

Cannonb4ll: I've tried that indeed, but then you'd need to remove it from the stub as it would return`null` otherwise. Certainly possible.

Let's wait what @reinink thinks?

claudiodekker: Yes, sure :) 

One thing I'd like to add: Personally, I think I'd more often than not use the `prop` instead of the entire method to set the root view.. It's a bit shorter/simpler, and like I already mentioned on Discord I personally haven't ever had the need to use logic to determine the root view, and unless I'm wrong, I think the same would apply to most other situations?

src/Middleware.php
```diff
@@ -129,4 +133,15 @@ public function resolveValidationErrors(Request $request)
             return $bags->has('default') ? $bags->get('default') : $bags->toArray();
         });
     }
+
+    /**
+     * Ability to set the current root view.
+     *
+     * @param Request $request
```
claudiodekker, line 140:
```suggestion
     * @see https://inertiajs.com/server-side-setup#root-template
     * @param Request $request
```

src/Middleware.php
claudiodekker: @Cannonb4ll I just briefly discussed this with @reinink, and he suggested to do pretty much exactly what I mentioned previously. I've implemented those changes, as well as added the tests that I mentioned I would add before.

src/Middleware.php
```diff
@@ -8,6 +8,14 @@
 
 class Middleware
 {
+    /**
+     * The root template that will be loaded on the first page visit.
```
reinink, line 12:
```suggestion
     * The root template that's loaded on the first page visit.
```

src/Middleware.php
```diff
@@ -42,6 +50,18 @@ public function share(Request $request)
         ];
     }
 
+    /**
+     * Sets the root template that will be loaded on the first page visit.
```
reinink, line 54:
```suggestion
     * Sets the root template that's loaded on the first page visit.
```

stubs/middleware.stub
```diff
@@ -7,6 +7,14 @@ use Inertia\Middleware;
 
 class {{ class }} extends Middleware
 {
+    /**
+     * The root template that will be loaded on the first page visit.
```
reinink, line 11:
```suggestion
     * The root template that's loaded on the first page visit.
```

PR #171, opened 2020-10-22
Add manifestPath property to middleware

hailwood:
Presently the manifest path is hardcoded,
This property allows users who either aren't using mix (e.g. using encore by symfony or a custom webpack setup) or have mix outputting to a non-root directory to just override this one property rather than duplicating the current manifest logic.

I've left the property out of the stub as I figure it's not a common enough use case to need it in the stub, and those that are running a custom build process are likely comfortable enough to take a quick peek at the parent version method.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -15,10 +15,13 @@ class Middleware
      * @var string
      */
     protected $rootView = 'app';
-    
+
     /**
-     * The path to the manifest file relative to the public directory.
-     * Used to determine the current asset version
+     * The path to the manifest file relative to the public directory
+     * used to determine the current asset version.
+     *
+     * @see version
+     * @var string
      */
     protected $manifestPath = 'mix-manifest.json';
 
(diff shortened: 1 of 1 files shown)


reinink:
Hey there @hailwood! Thanks for contributing this, but I don't think this really makes sense. The idea behind the default `version()` method is to provide a implementation that will work for most Laravel applications. However, if the two approaches we use there don't make sense, then that's exactly when it makes sense to just implement your own custom `version()` method in your middleware. Something like this:

```php

use Illuminate\Http\Request;
use Inertia\Middleware;

{
    /**
     * Determines the current asset version.
     *
     * @see https://inertiajs.com/asset-versioning
     * @param  \Illuminate\Http\Request  $request
     * @return string|null
     */
    public function version(Request $request)
    {
        return md5_file(public_path('your-manifest-file.whatever'));
    }
}
```

I hope that helps! 🤟

hailwood:
That's fair, 
That's exactly what I'm doing currently, it was an idea that popped into my head and since it look all of 5 min to PR figured why not 🙂

PR #178, opened 2020-10-29
Add optional id argument to the blade directive

daniel-werner:
Hi,

I had a situation where a div with id="app" already existed where I wanted to use the @inertia directive. 
I solved it by changing the other div, but I thought it would be good to be able to pass the id as an optional argument to the directive like @interia('div-id')

This PR implements the above mentioned feature, and adds a test to it.

Please let me know what do you think.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -24,8 +24,8 @@ class ServiceProvider extends BaseServiceProvider
 
     protected function registerBladeDirective()
     {
-        Blade::directive('inertia', function () {
-            return '<div id="app" data-page="{{ json_encode($page) }}"></div>';
+        Blade::directive('inertia', function ($id = '') {
+            return sprintf('<div id="%s" data-page="{{ json_encode($page) }}"></div>', $id ?: 'app');
         });
     }
 
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -25,6 +25,8 @@ class ServiceProvider extends BaseServiceProvider
     protected function registerBladeDirective()
     {
         Blade::directive('inertia', function ($id = '') {
+            $id = trim($id, '\'"');
+
             return sprintf('<div id="%s" data-page="{{ json_encode($page) }}"></div>', $id ?: 'app');
         });
     }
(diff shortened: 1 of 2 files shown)


claudiodekker:
Although Inertia is primarily intended to be used as the only 'real' thing on the page, I do like this addition 👍

claudiodekker:
@daniel-werner Did you test this manually as well? Apparently we had this feature before (https://github.com/inertiajs/inertia-laravel/issues/139), which caused (https://github.com/inertiajs/inertia-laravel/issues/140). 

Admittedly, that one didn't have any tests, but the implementation itself looks very similar?

daniel-werner:
@claudiodekker Yes, I just tested this manually and it works. 

However during the implementation and testing I also encountered the same issue from #140. 
It is a bit tricky, the reason of this problem is how blade handles custom directives (this code is from the blade compiler):
```
    protected function callCustomDirective($name, $value)
    {
        if (Str::startsWith($value, '(') && Str::endsWith($value, ')')) {
            $value = Str::substr($value, 1, -1);
        }

        return call_user_func($this->customDirectives[$name], trim($value));
    }
```

So, when the directive is called without arguments, an empty string is passed to the function, because of the `trim($value)`.

This is why the default argument value ($id = 'app') is not working because it is actually always called with an argument. So  checking for an empty value and defaulting to 'app' works.

Hope my explanation makes sense.

claudiodekker:
Managed to test this, and it indeed works mostly as you described:
- `@inertia
```<div id="app" data-page="...```

- `@inertia('test')
```<div id="'test'" data-page="...``` <-- note the nested quotes!

- `@inertia(test)
``` <div id="test" data-page="...``` <-- working as intended

**Comment by reinink** [comment]
Is there anyway we can remove the quotes when doing this?

```blade
@inertia('test')
```

daniel-werner:
I think, we could trim the single and double quotes in the directive function

reinink:
@daniel-werner Are you able to try and make that happen?

daniel-werner:
@reinink Fixed it

reinink:
Thanks! Let me do some testing with this. 👍

claudiodekker:
I'll be closing this PR, as we've actually implemented this as part of #339

PR #179, opened 2020-10-30
Add new lazy method to Facade's Docblock

mrmonat:
The lazy method is not "documented" on the Inertia Facade. This PR fixes that.


reinink:
Good catch, thanks!

PR #186, opened 2020-11-15
Customize blade directive to allow changing the div id

RobertBoes:
While reading through #141 and #178 I noticed the `$expression` that's passed by the blade compiler wasn't treated as a PHP expression, but rather as a string. I think this should be treated as an expression to follow blade standards and this will allow people to conditionally change the id like so `@inertia(true ? 'foo' : 'bar')`.

This means `@inertia(foo)` is not allowed, since that's not a valid expression. However, all directives mentioned in the [blade docs](https://laravel.com/docs/8.x/blade) have valid PHP as parameter for the blade directives (`'title'`, `count($records) === 1`) and for example `yield(content)` wouldn't work either since that doesn't contain a valid PHP expression.

The easiest solution is to output the expression in the directive directly to pass that to a method call, so the directive would return this `'<?php echo app(\MyClass::class)->render('.$expression.'); ?>'`. This works as expected, unfortunately Inertia also needs to output the `$page` variable which is passed as view data. I've tried to come up with a solution for this, but this is far from perfect, since I don't really like the usage of an anonymous class for this. Maybe someone can help me figure out a better solution, or this PR can be used to improve one of the other PR's

Oh and I used @claudiodekker's tests from #141, so I added you as co-author :)

### DIFF
--- a/src/BladeDirective.php
+++ b/src/BladeDirective.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Inertia;
+
+class BladeDirective
+{
+    public function render(string $id = 'app')
+    {
+        return new class($id) {
+            private $id;
+
+            private $page;
+
+            public function __construct(string $id)
+            {
+                $this->id = $id;
+            }
+
+            public function withPage($page)
+            {
+                $this->page = $page;
+
+                return $this;
+            }
+
+            public function __toString()
+            {
+                return '<div id="'.$this->id.'" data-page="'.e(json_encode($this->page)).'"></div>';
+            }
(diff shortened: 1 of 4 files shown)


claudiodekker:
Honestly, the anonymous class really isn't that bad in this context. 
Sure, it's not super pretty, but it's a perfectly valid use-case IMO. 👍

claudiodekker (approved):

claudiodekker:
I'll be closing this PR, as we've actually already re-implemented this as part of #339 👍

PR #205, opened 2021-01-08
Make Response macroable

binotaliu:
Make Response macroable would be useful when we need to set the same view data very often:

For example, I'm not using vue meta, so meta data need to pass into Blade:

```php
Inertia::render('Dashboard')
       ->withViewData('noindex', true);﻿
```

Would be nice if we can just add a `withNoIndex()` method:
```php
Inertia\Response::macro('withNoIndex', fn () => $this->withViewData('noindex', true));
```

then we can instead just:
```php
Inertia::render('Dashboard')->withNoIndex();
```


reinink:
Thanks for this contribution. This has been released: [v0.3.6](https://github.com/inertiajs/inertia-laravel/releases/tag/v0.3.6).

PR #206, opened 2021-01-10
Add render url

ajnsn:
This PR adds support for specifiying a custom render url which is returned by `Inertia::render` as `page.url`

When updating a resource, you do not always want to `Redirect::back()`. Instead render a component with prop(s) after the user has updated the resource.

```js
this.$inertia.patch('products/123'), data, {
    preserveState: true,
});
```
```php
return Inertia::render('Products', [
    'product' => $product,
]);
```
 
Currently Inertia will return `/products/123` as `page.url` as `Response` [is always using `$request->getRequestUri()`](https://github.com/inertiajs/inertia-laravel/blob/726b1e0d4fc399089dd29820a78e40d1c9aad2eb/src/Response.php#L92)

What if the url should be `/products`?

This PR allows you to specify the rendered url, e.g:

```php
return Inertia::render('Products', [
        'product' => $product,
    ])
    ->renderUrl('/products');
```
I believe that there a other use cases where you want to customize the rendered url as well.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -20,6 +20,7 @@ class Response implements Responsable
     protected $rootView;
     protected $version;
     protected $viewData = [];
+    protected $renderUrl;
 
     public function __construct($component, $props, $rootView = 'app', $version = null)
     {
@@ -47,6 +48,13 @@ class Response implements Responsable
         } else {
             $this->viewData[$key] = $value;
         }
+
+        return $this;
+    }
+
+    public function renderUrl($renderUrl)
+    {
+        $this->renderUrl = $renderUrl;
 
         return $this;
     }
@@ -89,7 +97,7 @@ class Response implements Responsable
         $page = [
             'component' => $this->component,
             'props' => $props,
-            'url' => $request->getRequestUri(),
+            'url' => $this->renderUrl ?? $request->getRequestUri(),
             'version' => $this->version,
         ];
 
--- a/tests/ResponseTest.php
+++ b/tests/ResponseTest.php
@@ -251,4 +251,22 @@ class ResponseTest extends TestCase
         $this->assertSame('Jonathan Reinink', $user['name']);
         $this->assertTrue($user['can']['deleteProducts']);
     }
+
+    public function test_can_set_render_url()
+    {
+        $request = Request::create('/products/123', 'PATCH');
+        $request->headers->add(['X-Inertia' => 'true']);
+
+        $product = (object) ['name' => 'My example product'];
+        $response = new Response('Products', ['product' => $product], 'app', '123');
+        $response = $response->renderUrl('/products');
+        $response = $response->toResponse($request);
+        $page = $response->getData();
+
+        $this->assertInstanceOf(JsonResponse::class, $response);
+        $this->assertSame('Products', $page->component);
+        $this->assertSame('My example product', $page->props->product->name);
+        $this->assertSame('/products', $page->url);
+        $this->assertSame('123', $page->version);
+    }
 }


darklord1807:
Thanks. Really needed this!

ajnsn:
@reinink @claudiodekker 
When you have time, could you take a look and review this PR? Thank you!

reinink:
Hey thanks for this PR @ajnsn and for your interest in Inertia.js.

Can you help me understand why a redirect to the products page (`/products`) isn't the right solution here? I agree that `Redirect::back()` isn't always the right solution (you may not want to go back to the edit product page on success), but you can redirect to other pages instead using `Redirect::to($url)`.

The huge benefit with doing a redirect is that you don't have to duplicate the logic to generate the products page in both the `ProductsController@index` method and the `ProductsController@update` method. Rather, you simply submit a patch request to `/products/123`, and then redirect to `/products` page on success. Something like this:

```php
public function update($product)
{
    $product->update(
        Request::validate([
            'name' => 'required',
        ])
    );

    return Redirect::to('/products');
}
```

Would that not work? What am I missing? 🤔

ajnsn:
Hey @reinink. 
thank you for taking time looking into this 👍 

Typically I pass back `props` to the original component and want to keep using Inertia. I  preseve the state when posting, patching, putting data, let the backend do the work and do something with the updated props coming back. 

**Example 1**
The updated record.

```php
return Inertia::render('products', [
        'updatedProduct' => ProductResource::make($product),
    ])
    ->renderUrl('/products');
```

This is very useful when the component already contains a lot of data and you do not want to reload the whole data when redirecting to the it's route. You just watch the `updatedProduct` and insert it into the original data.

**Example 2**
A command.

```php
return Inertia::render('products', [
        'isCreatingProduct' => false,
    ])
    ->renderUrl('/products');
```

When using modals to create/edit a resource, this is extremly helpful to just close the modal. I already wrote some words about this approach [in this thread](https://github.com/inertiajs/inertia/discussions/249#discussioncomment-225590).

Some other recent examples reported by other developers:
[#453](https://github.com/inertiajs/inertia/discussions/453#discussion-1931434)
[#487](https://github.com/inertiajs/inertia/discussions/487#discussion-3152706) 
[#468](https://github.com/inertiajs/inertia/discussions/468#discussioncomment-383672)

johanvanhelden:
To be honest, I am still a little confused as to what this solves. Would it also solve the following situation that occurred in one of my apps:

- There a page to edit an entity.
- On that page, the user can trigger a 2-step workflow to, for example, delete that entity. Step 1 is a confirmation screen and step 2 is the confirmation page to tell the user the entity has been deleted and contains a button to redirect the user to a different place in the application.

Using Inertia out of the box, after clicking the "Confirm deletion" button (step 1 of the workflow) I would like to just like to redirect back from the back-end, but this is no longer possible because the entity has been deleted. This would trigger a 404-page. I also can not redirect the user to a different page, because I want to show the user the final confirmation step of the modal. So I want the user to remain on the same page, but I also want some props to refresh with the up-2-date information.

ajnsn:
> I would like to just like to redirect back from the back-end

@johanvanhelden In this case I would use the approach that @reinink suggested [above](https://github.com/inertiajs/inertia-laravel/pull/206#issuecomment-783359760). Just `Redirect::to` to the list of your "entities". You [may also pass session data](https://laravel.com/docs/8.x/redirects#redirecting-with-flashed-session-data), e.g.

```php
 return Redirect::to('/products')->with('successMessage', 'This product has been deleted');
```

You can [pass this data via shared props to your frontend](https://inertiajs.com/shared-data#flash-messages) and show the confirmation message there. This could also work as an universal approach to display notifications across your apps.

johanvanhelden:
The problem was that the designer really designed the last step inside of the modal. So redirecting to the list of entities with a confirmation message was not an option. I had to stay inside of the page with the modal.

ajnsn:
@johanvanhelden Have a look at the [event callbacks](https://inertiajs.com/manual-visits#event-callbacks). A lot of things can be realized inside the frontend. If there are further questions, I recommend [asking / starting a thread here](https://github.com/inertiajs/inertia/discussions).

ahmedkandel:
Hello, @reinink & @ajnsn

I have a use case that may benefit from this PR.

The index page is a DataTable with (filters, sort, pagination, ...) to lazy load a subset of users. Those parameters are always sent to the controller index method.

When opening a modal to edit a user then post/patch it to the backend I also post those DataTable parameters to sync the state with the backend.

If i follow @reinink [approach ](https://github.com/inertiajs/inertia-laravel/pull/206#issuecomment-783359760) the redirect will not pipe the $request from update method to index method. So i simply return index($request) method from the update($request) method but the page.url gets changed to the update route url '/user/{user}'.

That is why this PR may be a solution.

### Route:
``` php
Route::match(['GET', 'POST'], '/users', [UserController::class, 'index'])->name('users.index');

Route::patch('/user/{user}', [UserController::class, 'update'])->name('users.update');
```

### Controller:
``` php
{
    $query = User::where('active', true);

    // Filter

    // Sort

    // Pagination

        'users' => $query->get(),
    ]);
}

public function update(User $user, Request $request)
{
    // Updating
    
    $user->save();

    return $this->index($request);
}
```

Thanks

reinink:
@ahmedkandel So, I assume that these filters and stuff are set as `GET` parameters in your app, right? Something like `/users?sort=name&page=2`? If that's the case, doing a `Redirect::back()` within your `update()` method will preserve these query strings automatically.

ahmedkandel:
Thanks, @reinink those filters are set as `POST` as they contain arrays such as constraints as some complicated logic.

reinink:
@ahmedkandel So, I'm not telling you how to build your apps (😅), but I personally think using `POST` for search filters, sorting and pagination isn't really the right solution, and as a side affect, it's causing this data reload issue for you. In my experience, you can generally make this work with a GET request and query strings. You can absolutely include arrays in query strings. Alternatively (and I've done this before), you could create a JSON object, and then pass that as a query string. Something like `/users?query=YOUR_JSON_OBJECT`.

ahmedkandel:
I may use Laravel `withInput()` while redirecting but it will make the controller index method more complex:

### Update method
```php
return redirect()->route('users.index')->withInput();
```

### Index method
```php
{
    $query = User::where('active', true);

    // Filter
    if ($request->filters || $request->old('filters'))

    // Sort

    // Pagination

        'users' => $query->get(),
    ]);
}
```

ahmedkandel:
@reinink First thanks for your time.

Although Using `GET` request will generate an ugly long URL but makes sense :ok_hand:

I have wraped all the (filters, sort, pagination, ....) in an object and passed it to `Inertia.get(route('users.index'), params)`.

But I faced another strange issue as the params object has nested arrays, so when it gets stringified by Inertia `mergeDataIntoQueryString` all the numerical indexes are lost and I couldn't reconstruct the object at the backend.

**Here is a piece of the URL params: []**
```
filters[constraints][][matchMode]=startsWith&filters[constraints][][value]=a&filters[constraints][][matchMode]=endsWith&filters[constraints][][value]=z
```
**At the backend will generate:**
```
array:4 [
  0 => array:1 [
    "matchMode" => "startsWith"
  ]
  1 => array:1 [
    "value" => "a"
  ]
  2 => array:1 [
    "matchMode" => "endsWith"
  ]
  3 => array:1 [
    "value" => "z"
  ]
]
```

**While it should be: [0]**
```
filters[constraints][0][matchMode]=startsWith&filters[constraints][0][value]=a&filters[constraints][1][matchMode]=endsWith&filters[constraints][1][value]=z
```
**To generate:**
```
array:2 [
  0 => array:2 [
    "value" => "a"
    "matchMode" => "startsWith"
  ]
  1 => array:2 [
    "value" => "z"
    "matchMode" => "endsWith"
  ]
]
```

After a little digging, I see that Inertia is using `qs` with [`arrayFormat: 'brackets'`](https://github.com/inertiajs/inertia/blob/896c53bb0e80bb6ba56e588a3afe665fa8aa0498/packages/inertia/src/url.js#L13), Do you think this needs to be [`arrayFormat: 'indices'`](https://github.com/ljharb/qs) ?

darklord1807:
@reinink  I think it is a good thing to add it as an option.  Eg. I am using this when calling different functions in the main controller to split up the different codes. Without this feature calling a function like "Add Record" in the main controller would need to reside inside the Index function which makes things complicated as the project grow in size. 

I do not think it would affect much if we enable this feature. 

Thanks for your help

ajnsn:
fyi if anyone want to solve this temporarly for now could use a Macro and a custom Response class to "override" the current `render` method. Sample:

 InertiaMacroServiceProvider

```php
namespace App\Providers;

use App\Inertia\Response;
use Illuminate\Support\ServiceProvider;
use Inertia\ResponseFactory;

class InertiaMacroServiceProvider extends ServiceProvider
{
    public function boot()
    {
        ResponseFactory::macro('renderWithUrl', function (string $component, array $props = []) {
            if ($props instanceof Arrayable) {
                $props = $props->toArray();
            }

            return new Response(
                $component,
                array_merge($this->sharedProps, $props),
                $this->rootView,
                $this->getVersion()
            );
        });
    }
}
```

Response

```php

namespace App\Inertia;

use Illuminate\Http\JsonResponse;
use Inertia\Response as InertiaResponse;

class Response extends InertiaResponse
{
    protected $renderUrl;

    public function url($renderUrl)
    {
        $this->renderUrl = $renderUrl;

        return $this;
    }

    {
        $response = parent::toResponse($request);

        if ($response instanceof JsonResponse) {
            $currentData = $response->getData();

            $currentData->url = $this->renderUrl ?? $currentData->url;

            $response->setData($currentData);
        }

        return $response;
    }
}
```php

Usage:

```php
return Inertia::renderWithUrl('Products', [
        'product' => $product,
    ])
    ->url('/products');
```

Wulfheart:
Any plans to include this or is it closed definitely?

Wulfheart:
But this causes the url bar to flicker a little bit. It don’t like this. Any way to mitigate this?

PR #231, opened 2021-03-11
Fix casing of X-Inertia-Error-Bag HTTP header

szepeviktor:
As per https://github.com/inertiajs/inertia/blob/896c53bb0e80bb6ba56e588a3afe665fa8aa0498/packages/inertia/src/inertia.js#L231

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -148,8 +148,8 @@ class Middleware
                 return $errors[0];
             })->toArray();
         })->pipe(function ($bags) use ($request) {
-            if ($bags->has('default') && $request->header('x-inertia-error-bag')) {
-                return [$request->header('x-inertia-error-bag') => $bags->get('default')];
+            if ($bags->has('default') && $request->header('X-Inertia-Error-Bag')) {
+                return [$request->header('X-Inertia-Error-Bag') => $bags->get('default')];
             } elseif ($bags->has('default')) {
                 return $bags->get('default');
             } else {


reinink:
Did you actually run into a bug with this? The reason we're using lowercase headers in PHP (Laravel) is because the `Symfony\Component\HttpFoundation\HeaderBag` automatically lowercases them.

```php
dd($request->headers);
```

image

szepeviktor:
> `Symfony\Component\HttpFoundation\HeaderBag` automatically lowercases them.

Isn't HTTP/2 _who_ lowercases them??

szepeviktor:
No!

https://github.com/symfony/http-foundation/blob/5.x/HeaderBag.php#L131

Symfony does it.

May I lowercase all other header names?

X-Inertia
X-Inertia-Version
X-Inertia-Partial-Data
X-Inertia-Partial-Component

szepeviktor:
@reinink What do you think?

reinink:
> May I lowercase all other header names?

I'm not sure what you mean. These headers are set by Inertia, and then server-side in PHP they are lowercased. This all just works right now, so I'm not really sure what you're asking.

I'd suggest we just leave things as they are, unless there's a bug.

szepeviktor:
> I'd suggest we just leave things as they are, unless there's a bug.

When I've received the header list from a developer using Inertia I've noticed one of them is all lowercase.

@reinink Could you we make all of them the same case?

PR #244, opened 2021-04-02
Add flushShared() method for clearing all shared data

reinink:
This PR adds a new `Inertia::flushShared()` method, which can be used to completely clear all the shared data.


reinink:
Released: https://inertiajs.com/releases/inertia-laravel-0.4.1

PR #252, opened 2021-04-11
Fixing methods getting ignored after redirect

hicham-saddek:
### Subject

The inertia middleware ignores the methods comming from the vuejs to the laravel application, after they are redirected many times, they are all treated as `GET` methods.

### How to reproduce
 1- Fresh Laravel application with Jetstream installed.
 2- If cPanel available, activated the automatic `HTTP` to `HTTPS` redirect.
 3- Try activating the Two-Factor-Authentication via the Frontend.

### Solution

Since the redirected method is actually ignored but is represented in the header `REDIRECT_REDIRECT_REQUEST_METHOD` as the redirect method, we can actually override the current method, with the redirect one if it's existing.

which we can find in ``$request->server('REDIRECT_REDIRECT_REQUEST_METHOD')``.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -158,7 +158,7 @@ class Middleware
             }
         });
     }
-    
+
     /**
      * Resolves and prepares the request methods, if they are being redirected.
      *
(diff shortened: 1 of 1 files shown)


hicham-saddek:
@reinink any updates ?

hicham-saddek (commented):
Grammar fix needed

driesvints:

PR #262, opened 2021-05-10
Register macros on `register()` to avoid race condition

crynobone:
Normally all routes will be registered during `boot()`, if a package service provider using Inertia get loaded first before `Inertia\ServiceProvider` they will face the following error:

```
InvalidArgumentException: Attribute [inertia] does not exist.
```

-----

Since `macro` uses static function it should be safe to register it earlier.


claudiodekker:
Thanks Mior! 👍

PR #265, opened 2021-05-12
Use array for responsable getData()

rubenvanassche:
At the moment when a `Responsable` prop is transformed, it is transformed to an object:

```php
if ($prop instanceof Responsable) {
    $prop = $prop->toResponse($request)->getData();
}
```

A PHP object has a few limitations, when it comes to naming keys. For example, this is valid json:

```json
{"\u0000*\u0000_invalid_key": "for object"}
```

It is also a valid PHP array:

```
[
  "\x00*\x00_invalid_key" => "for object"
]
```

But it is not a valid PHP object, `json_decode` will return `null`.

This PR changes how a `Responsable` is transformed, it will not transform it to an object like it is now but it will transform it to an `array` by changing the above code to this:

```php
if ($prop instanceof Responsable) {
    $prop = $prop->toResponse($request)->getData(true);
}
```

I've also included a test which showcases this example.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -71,7 +71,7 @@ class Response implements Responsable
             }
 
             if ($prop instanceof Responsable) {
-                $prop = $prop->toResponse($request)->getData();
+                $prop = $prop->toResponse($request)->getData(true);
             }
 
             if ($prop instanceof Arrayable) {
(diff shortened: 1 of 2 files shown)


reinink:
Hmm, does this have the potential of breaking any existing code? Does it change the final outputted JSON at all?

rubenvanassche:
Hi @reinink,

Sorry for the late response, kinda forgot about this one 🙈 . Since the object in `$prop` is later converted into JSON it doesn't matter if we convert it to an array first. Only these strange keys will be converted to JSON whereas they now let the application crash.

claudiodekker:
Thanks for this @rubenvanassche! 👍

PR #276, opened 2021-06-13
Does PHP 7.2 support return typehints?

szepeviktor:
❓


szepeviktor:
Yes.
https://github.com/szepeviktor/inertia-laravel/pull/1/checks

PR #280, opened 2021-06-24
Ability to chain the root view

MarGul:
Very small change and mostly syntax-sugar.

I have a guest section of my app where I need a new root-view.

Instead of doing 
```
Inertia::setRootview('guest');

return Inertia::render('Guest/Home');
```

I can now do this instead: `return Inertia::setRootView('guest')->render('Guest/Home');`

Nothing much but I like it this way instead. What do you say?

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -51,6 +51,13 @@ class Response implements Responsable
         return $this;
     }
 
+    public function rootView($rootView)
+    {
+        $this->rootView = $rootView;
+
+        return $this;
+    }
+
     public function toResponse($request)
     {
         $only = array_filter(explode(',', $request->header('X-Inertia-Partial-Data')));


claudiodekker:
Hey @MarGul,

Thanks for this! I do dig the idea, although the only thing I somewhat dislike is that:
- `setRootView` doesn't read that nice when chaining
- People rely on the current implementation of `setRootView`

Which makes me wonder whether we should (in addition to this change) add an 'alias' method that we could use for this? In either case, I'm definitely OK with these changes, so I'll just approve it for now as-is.

@reinink thoughts?

claudiodekker (approved):

MarGul:
@claudiodekker agree that the name `setRootView` doesn't read very well when chaining. Maybe we could add the rootview as a third param in `render`? Like `Inertia::render('Gest\Home', ['param' => 'test'], 'guest')`? Basically "just" make the render method call `setRootView` before returning the response? Maybe `render` is doing too much then but just an idea.

reinink:
Hmm, my primary concern with this change is that `Inertia::setRootView()` currently affects things globally for Inertia. It will change the root view for all subsequent Inertia responses. I worry that chaining like this makes it feel like a local change only. For example:

```php
$authResponse = Inertia::render('Auth/Home');
$guestResponse = Inertia::setRootView('guest')->render('Guest/Home');

if (Auth::user()) {
    return $authResponse; // will use "guest" root view
}
```

Obviously this is silly code, but it illustrates that someone could inadvertently change the root view globally without realizing it.

The better way to solve this is not in the `ResponseFactory`, but rather in the `Response` class. That way the root view change would just be applied to that response, and not changed globally. It also solves the naming issue.

```php
return Inertia::render('Guest/Home')->rootView('guest');
```

Feel free to update this PR accordingly. 👍

reinink (changes requested):
Add `rootView()` to the `Response` class instead, as per [this comment](https://github.com/inertiajs/inertia-laravel/pull/280#issuecomment-888211336).

MarGul:
This was a way better solution @reinink ! I have pushed the code now and tried it out. Works perfect for me now because I have one rootView for my `guest` area and one rootView for my `admin` area that I easily can handle now. Thanks!

reinink:
@MarGul Awesome, thanks!

reinink:
This has been released: https://github.com/inertiajs/inertia-laravel/releases/tag/v0.4.4

(I still need to write the release notes)

dillingham:
@reinink Maybe for 1.0 it could be called `layout()` throughout?

Nice feature for sure 🙌

PR #286, opened 2021-07-31
Add Vite manifest to default version check

innocenzi:
In the same spirit as https://github.com/inertiajs/inertia-laravel/pull/285, I'm trying to reduce friction for Vite users by enabling version checks by default on the Inertia side. 

This PR adds a fallback check for `public/build/manifest.json`, the same way the Mix manifest is checked. This path is the default for anyone using [Sebastian De Dayne's guide](https://sebastiandedeyne.com/vite-with-laravel/) or [Laravel Vite](https://laravel-vite.innocenzi.dev/guide/production.html#building-for-production). 

Notes:
- I think this change is good because this would be a "gotcha" in the sense that this functionality is easily forgotten (especially by newcomers), and that this doesn't cause a behavior change in any way
- I could've used an array instead of copying the three lines but I didn't want to add unnecessary complexity to the code
- Web manifests, usually `app.webmanifest` or `manifest.json`, are at the root of the public directory so they won't conflict

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -32,6 +32,10 @@ class Middleware
         if (file_exists($manifest = public_path('mix-manifest.json'))) {
             return md5_file($manifest);
         }
+        
+        if (file_exists($manifest = public_path('build/manifest.json'))) {
+            return md5_file($manifest);
+        }
     }
 
     /**


innocenzi:
Other StyleCI issues are unrelated to this PR. We might want to do a StyleCI pass and merge it directly.

claudiodekker:
I'm curious regarding the path itself, is 'public/build' that common, or is it changeable with Vite?

innocenzi:
> I'm curious regarding the path itself, is 'public/build' that common

As far as I remember I have always seen `build` being used as the directory that host build files. One example I can quote is [Webpack Encore](https://symfony.com/doc/current/frontend/encore/simple-example.html), which is the first Webpack abstraction I have used. 

When using Mix for the first time I thought it was weird to disperse everything in `public/js` or `public/css`, so I always modified that to be in a `build` directory anyway. 

I've also seen `/assets` being used (by Adonis and it's also Vite's default, notably).

> or is it changeable with Vite?

It's definitely changeable. When using `laravel-vite`, you just need to change the `vite.build_path` option, for instance.

PR #289, opened 2021-08-10
Check if $errors index 0 exists

patrasq:
**Context:**
I've created a helper in order to handle errors by class.
For some reasons, even though index 0 existed in $errors, the returned response was:
`Undefined array key 0 in file inertiajs\inertia-laravel\src\Middleware.php on line 148`

After using this fix no problems were encoutered.

**Fix:**
Added a simple check to see if index 0 exists in $errors.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -145,7 +145,7 @@ class Middleware
 
         return (object) collect($request->session()->get('errors')->getBags())->map(function ($bag) {
             return (object) collect($bag->messages())->map(function ($errors) {
-                return $errors[0];
+                return array_key_exists(0, $errors) ? $errors[0] : false;
             })->toArray();
         })->pipe(function ($bags) use ($request) {
             if ($bags->has('default') && $request->header('x-inertia-error-bag')) {


claudiodekker:
Hey @patrasq 👋,

Could you share a clear and reproducable example of a situation in which this problem occurs? I'd love to merge this in, but because nobody else has ran into this yet and the problem itself doesn't seem that clear I'm holding off on this for now.

Thanks!

ConnorMoodyDev:
@claudiodekker I ran into this issue with a project that has been working fine up until recently. 

A simple inertia.post in vue to my registration route causes this issue whether the route returns anything or not. 
I commented out my entire controller function and the issues till occured.

ConnorMoodyDev:
Some more details on this:

The route this was occurring on was my Register controller.
The cause of it was this piece of code:

`
 public function __construct()
{
    $this->middleware('guest');
}

`

Going to the controller directly in the web browser caused the issue as well.

reinink:
Hey folks, I am going to close this, as we have not been able to reproduce it. If someone can provide a simple reproduction repo for this I'm happy to have another look. 👍

djoey123:
Hi,
Maybe late than never :P

How i can reproduce this issue was:
`return redirect()->route('index')->withErrors(['error' => null]);`

My code was like this.
`        try {
            $user = Socialite::driver('discord')->user();
        } catch (\Throwable $exception) {
            return redirect()->route('index')->withErrors(['error' => $request->get('error_description')]);
        }`
       
error_description is sent back by the socialite driver. but on the InvalidStateException no error_description was provided. So the default value is null.
Result was: Undefined array key 0 on line 146 inertia-laravel/Middleware.php

Hopefully some good information for the issue @reinink :)

PR #294, opened 2021-08-19
feat: trigger event before sending response

jaulz:
This PR will trigger an event before the actual Inertia response it sent. The name of the event is up to discussion and maybe you have a better idea. Having this event it will be easy to implement a widget for the Debugbar to inspect the props.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -101,7 +101,9 @@ class Response implements Responsable
             'version' => $this->version,
         ];
 
-        Event::dispatch('inertia.collected', [$page]);
+        if (env('APP_DEBUG', false)) {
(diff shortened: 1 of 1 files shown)


src/Response.php
```diff
@@ -100,6 +101,10 @@ public function toResponse($request)
             'version' => $this->version,
         ];
 
+        if (env('APP_DEBUG', false)) {
```
dillingham, line 104:
```php
config('app.debug')
```
dont use env outside of config files

jaulz: good hint, thanks!

MinePlay96:
I would be happy about it too

driesvints:

PR #302, opened 2021-09-19
Allow ResponseFactory location to be called with a redirect response

rojtjo:
This change allows the `Inertia::location()` method to accept a `RedirectResponse`.

In an app I'm developing using multiple subdomains I ran into the issue that I needed an external redirect after the authentication. Since I want to keep using the `redirect()->intended()` feature it feels natural to just pass in the whole `RedirectResponse` in the `location` method instead of manually getting the target url from the redirect.

Usage example:
```php
// Some controller action

return Inertia::location(
    redirect()->intended(RouteServiceProvider::HOME)
);
```

Please note that I'm using Symfony's `RedirectResponse` since Laravel's implementation just extends it.

Let me know what you think!

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -10,7 +10,6 @@ use Illuminate\Support\Facades\App;
 use Illuminate\Support\Facades\Request;
 use Illuminate\Support\Facades\Response as BaseResponse;
 use Illuminate\Support\Traits\Macroable;
-use Symfony\Component\HttpFoundation\RedirectResponse;
 
 class ResponseFactory
 {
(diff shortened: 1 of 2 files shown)


RobertBoes:
Another really useful benefit of this would be packages like Socialite. Currently you'll have to do something like the following to make this work:
```php
Route::get('/auth/github', function () {
  return Inertia::location(
    Socialite::driver('github')->redirect()->getTargetUrl()
  );
});
```
Allowing a `RedirectResponse` cleans this up nicely and makes the DX a lot better imo.

claudiodekker:
@rojtjo Correct me if I'm wrong, but my understanding is that you're redirecting to a non-inertia page in your example, right?

rojtjo:
Yes, you are correct. It's only purpose is to save some keystrokes 😅

RobertBoes:
@claudiodekker in the PR he mentioned he's using Inertia on multiple subdomains, but inertia doesn't work cross-domain (mostly because of the usage of the History API), so you'll have to force a visit to the new subdomain. So it's not only useful for non-Inertia pages, but also to go to another domain that does render an Inertia view

claudiodekker:
Thanks a lot, @rojtjo!

Issue #295, opened 2021-08-26
Potential `Inertia::location()` enhancement?


daronspence:
Heya! 👋 

If I'm `#doingitwrong` please let me know, but I ran into something that I think can be handled at the adapter level and wanted to run it by the team here before making a PR.

I have a controller method, where I want to hard reload the page after doing something. `Inertia::location` is a perfect use case, but it fails in the scenario where the user somehow ends up on that URL via a manual GET/POST request. In that scenario, the user gets a blank screen as the response is a 409. I know this is edge-case-y.

My proposed solution, is to check if the request is an Inertia request before sending the 409 response, otherwise send a normal redirect.

This would be the new method

```php
public function location($url)
{
  if (Request::inertia()) {
    return BaseResponse::make('', 409, ['X-Inertia-Location' => $url]);
  }
  return redirect($url);
}
```

If this has been covered elsewhere, apologies! Thanks again for making such a fantastic tool ❤️

innocenzi:
I've completely been bitten by that. Here's a snippet in my current codebase: 

```php
if (request()->inertia()) {
    return inertia()->location(config('legacy.login_redirect_url'));
}

return redirect()->to(config('legacy.login_redirect_url'));
```

PR #312, opened 2021-10-19
fix!: handle non-inertia `location` redirects

innocenzi:
Sometimes, we want to perform an external redirect that can be initiated on the first visit or on an Inertia visit.

In such a case, Inertia's `location` won't work because it will only simply return a normal `409` response with the `X-Inertia-Location`, but since the front-end is not loaded the end result will be a blank page. 

To fix this, we need to check if the current request is an Inertia one, perform the `location` redirect, or return a classic `RedirectResponse` otherwise. 

This PR fixes that issue by performing that check internally so the developer doesn't have to.

This is *technically* a breaking change since I had to modify the original test, but it's a breaking change that is more likely to fix issues than to create some.

---

TL;DR: with this PR, instead of doing:

```php
if (request()->inertia()) {
    return inertia()->location($url);
}

return redirect()->to($url);
```

We can just do:
```php
return inertia()->location($url);
```

---

Fixes https://github.com/inertiajs/inertia-laravel/issues/295

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -4,8 +4,10 @@ namespace Inertia;
 
 use Closure;
 use Illuminate\Contracts\Support\Arrayable;
+use Illuminate\Http\RedirectResponse;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\App;
+use Illuminate\Support\Facades\Request;
 use Illuminate\Support\Facades\Response as BaseResponse;
 use Illuminate\Support\Traits\Macroable;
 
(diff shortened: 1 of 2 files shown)


narasimhajupally:
Hello, I just checked the code changes for this issue and tried to do similar in my controller
```
if (Request::inertia()) {
       return Inertia::location(tenantRoute($email->domain->domain, 'home'));
}
return redirect(tenantRoute($email->domain->domain, 'home'));
```
I am getting this exception
`ErrorException Cannot unbind $this of closure using $this`
any idea ? 
Thanks.

claudiodekker:
Thanks @innocenzi!

PR #316, opened 2021-10-27
Add support for guzzle promises in props

rhysemmerson:
When making http requests to construct a props array a big performance gain can be achieved by making the requests asynchronously. 

Having support for this behaviour in the package will make this a seamless experience.

# Demonstration

As an example take two controller that hit this api route using laravel Http facade

```PHP
Route::get('/{foo?}', function ($foo) {
    sleep(1);
    return new JsonResource([
        'value' => $foo,
    ]);
});
```

## Control: synchronous example

```PHP
use Illuminate\Http\Client\PendingRequest;
use Inertia\Inertia;

class WelcomeController extends Controller
{
    private PendingRequest $client;

    public function __construct(PendingRequest $client)
    {
        $this->client = $client;

        Inertia::share([
            'one' => $this->client->get('/api/one'),
        ]);
    }

    public function index()
    {
        return Inertia::render('Welcome', [
            'two' => $this->client->get('/api/two'),
            'three' => $this->client->get('/api/three'),
        ]);
    }
}
```

Screen Shot 2021-10-27 at 5 54 38 pm

## Asynchronous Example

```PHP
use Illuminate\Http\Client\PendingRequest;
use Inertia\Inertia;

class WelcomeControllerAsync extends Controller
{
    private PendingRequest $client;

    public function __construct(PendingRequest $client)
    {
        $this->client = $client->async();

        Inertia::share([
            'one' => $this->client->get('/api/one'),
        ]);
    }

    public function index()
    {
        return Inertia::render('Welcome', [
            'two' => $this->client->get('/api/two'),
            'three' => $this->client->get('/api/three'),
        ]);
    }
}
```

Screen Shot 2021-10-27 at 6 01 01 pm

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,13 +27,13 @@
     "require": {
         "php": "^7.2|^8.0",
         "ext-json": "*",
-        "laravel/framework": "^5.4|^6.0|^7.0|^8.0",
-        "guzzlehttp/promises": "^1.5"
+        "laravel/framework": "^5.4|^6.0|^7.0|^8.0"
     },
     "require-dev": {
         "roave/security-advisories": "dev-master",
         "orchestra/testbench": "^4.0|^5.0|^6.0",
-        "phpunit/phpunit": "^8.0|^9.0"
+        "phpunit/phpunit": "^8.0|^9.0",
+        "guzzlehttp/promises": "^1.5"
     },
     "extra": {
         "laravel": {
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
     "require": {
         "php": "^7.2|^8.0",
         "ext-json": "*",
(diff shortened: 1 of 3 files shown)


rhysemmerson:
@reinink any thoughts on this- if this doesn't belong in the package is there some other way we can make it possible? 

I'm refactoring a http heavy app to make asynchronous requests. I can make it work by adding an abstraction on top of inertia but it would be great if all I had to do was return a promise.

claudiodekker:
Hey @rhysemmerson,

While I don't think we want a new dependency added to this package, I do think you can still keep the `use ...` and `instanceof ...` checks without any errors if they don't actually exist / aren't installed. That way, your code should still work, while others don't run into any issues.

Apart from that, this PR looks good to me.

rhysemmerson:
Thanks @claudiodekker I've removed the dependency

Issue #303, opened 2021-09-20
Laravel Passport authorization flow not working with Inertia?


abbluiz:
When Laravel Passport redirects back to an external URI after the OAuth 2.0 authorization flow, it renders an iframe of the external website, instead of actually redirecting to the external website.

abbluiz:
The thing is I could use `Inertia::location($url)` to redirect to the external client URL, but this is internal of Passport. The workaround I found was to override the handle method in `HandleInertiaRequests.php`:

```php
    public function handle(Request $request, Closure $next)
    {
        $parentResponse = parent::handle($request, $next);

        if (!empty($request->redirect_uri)) {
            return Inertia::location($parentResponse->headers->all()['location'][0]);
        }

        return $parentResponse;
    }
```

abbluiz:
I have updated my middleware. It is easier to understand now:
```php
public function handle(Request $request, Closure $next)
{
    $parentResponse = parent::handle($request, $next);

    $location = $parentResponse->headers->get('Location');

    if ($this->isExternal($location)) {
        if (request()->inertia()) {
            return inertia()->location($location);
        }

        return redirect()->to($location);
    }

    return $parentResponse;
}
```
A method to check if an URL is external, which is used above:
```php
public function isExternal($url)
{
    $internalHost = parse_url(config('app.url'), PHP_URL_HOST);
    $internalPort = parse_url(config('app.url'), PHP_URL_PORT);

    $host = parse_url($url, PHP_URL_HOST);
    $port = parse_url($url, PHP_URL_PORT);

    return $host !== $internalHost || $port !== $internalPort;
}
```

claudiodekker:
Hey @abbluiz, thanks for reporting this issue.

Whether Inertia is compatible with Passport or any third-party library really depends on the third-party library for the most part, as Inertia in and by itself is just a generic library without. However, I'll leave this issue open so that we can investigate and perhaps come up with a fix for this specific to Passport later.

My personal approach to this would probably be to add a special route that uses `Inertia::location()` to redirect to the Passport OAuth route. That way, you immediately (as the first step in the chain) tell Inertia that we're no longer using Inertia going forward, allowing Passport to properly render it's routes/do it's thing. Then afterwards, you can make sure it redirects back to your Inertia pages.

jessarcher:
Hey there,

We're closing this issue because it's inactive, already solved, old or not relevant anymore. Feel free to open up a new issue if you're still experiencing this problem.

[...]

PR #323, opened 2021-11-26
Fix CORS Issue with Laravel Passport

rogervila:
Hello,

When using Laravel Passport on Jetstream powered with Inertia, I get a CORS issue because the callback URL is handled by Inertia with an AJAX Request.

image

In order to solve it, I have modified my `HandleInertiaRequests` middleware to detect external locations and force redirections without AJAX.

### DIFF
--- a/stubs/middleware.stub
+++ b/stubs/middleware.stub
@@ -2,8 +2,11 @@
 
 namespace {{ namespace }};
 
+use Closure;
 use Illuminate\Http\Request;
+use Illuminate\Support\Str;
 use Inertia\Middleware;
+use Inertia\Inertia;
 
 class {{ class }} extends Middleware
 {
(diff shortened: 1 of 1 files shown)


claudiodekker:
Hi @rogervila,

This looks like an issue with the CORS headers not being present on the back-end routes. Because this is more of a Laravel specific issue that can be solved using it's CORS middleware & config and not something that's specific to Inertia, I'll be closing this PR.

More info:
- https://laravel.com/docs/8.x/routing#cors
- https://laravel.com/docs/8.x/sanctum#cors-and-cookies

rogervila:
Hi @claudiodekker is ok to close it then, I will review CORS headers.

abbluiz:
@rogervila Have you figured out the CORS settings for this to work?

rogervila:
@abbluiz not yet. 

I have tested with 2 Laravel projects created from scratch without modifications:

When Passport is installed on a plain Laravel project with blade views, it works. 

The same process with a Laravel project with Jetstream and Inertia fails because the login form submission triggers an AJAX request to the client site instead of making a full redirection. 

This is why I implemented this change, to make sure that outgoing redirections trigger a full reload.

abbluiz:
@rogervila I have managed to fix the CORS problem with Passport, but the issue with cors was really with the OAuth 2.0 client (which was a Next.js application in my case).

But, I still had to make similar modifications to make proper redirects. Check this issue https://github.com/inertiajs/inertia-laravel/issues/303 when you get a chance.

PR #333, opened 2021-12-16
Fix page.url doesn't contain url prefix

ppodds:
[Laravel PR 40014](https://github.com/laravel/framework/pull/40014)

Same issue. This PR fixes the URL prefix problem in `src/Response.php`. Making us deploy Laravel + Inertia project in a subfolder of domain possible.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -96,7 +96,7 @@ class Response implements Responsable
         $page = [
             'component' => $this->component,
             'props' => $props,
-            'url' => $request->getRequestUri(),
+            'url' => $request->getBaseUrl().$request->getRequestUri(),
             'version' => $this->version,
         ];
 


claudiodekker:
Hey 👋,

Thanks for your pull request to Inertia!

Could you please provide additional information that includes an explanation of the exact problem, how to reproduce said problem, and how this change fixes that problem? It is difficult for us to consider pull requests without this, as it otherwise may introduce breaking changes to the library. 

Thanks!

ppodds:
# Problem Introduce
Inertia generates the wrong `url` when the Laravel project deploys under the subfolder of the domain.

# How To Reproduce
This problem only occurs when you deploy your Laravel project in a subfolder of the domain like this.

`https://my.domain/to/project`

NGINX setting (reverse proxy) maybe like this

```
server {
  listen 80;
  listen [::]:80;

  server_name _;
  location /to/project/ {
    proxy_set_header X-Forwarded-Prefix /to/project;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
    proxy_set_header X-Forwarded-Host my.domain;
    proxy_set_header X-Forwarded-Port 443;
    # laravel backend at localhost:8080
    proxy_pass http://localhost:8080/;
}
```

In the said condition below, inertia will set `url` = `` when user requests `https://my.domain/to/project`, cause user's browser URL to be rewritten as `https://my.domain/`.

# Fix

`toResponse($request)` (in `src/Response.php`) doesn't consider base URL in `$page.url`. This PR just add base URL in front of request URI to fix it.

```php
<?php
...
class Response implements Responsable
{
    use Macroable;
    ...

    {
        ...
        $page = [
            'props' => $props,
            'url' => $request->getBaseUrl().$request->getRequestUri(),
            'version' => $this->version,
        ];
        ...
    }
}
```

claudiodekker:
Thanks!

PR #220, opened 2021-03-02
Merge inertia-laravel-testing into inertia-laravel

claudiodekker:
# Testing
After multiple months of beta testing, this PR makes [the inertia-laravel-testing library](https://github.com/claudiodekker/inertia-laravel-testing) an official part of the inertia-laravel adapter 🎉 

While the inertia-laravel-testing library WILL remain available for installation, but WILL NOT receive any further (security) updates going forward.

# Migrating

To migrate from `inertia-laravel-testing` to the `inertia-laravel` adapter, simply follow these easy steps:
1. In your `composer.json` file, remove the `claudiodekker/inertia-laravel-testing` dependency line
2. In your `composer.json` file, change the version of `inertiajs/inertia-laravel` to `^0.4.0`
3. Run `composer update`
4. Search and replace `use ClaudioDekker\Inertia\Assert;` with `use Inertia\Testing\Assert;` in your test suites.

That's it! When done correctly, all of your tests should still be passing.

PR #338, opened 2021-12-26
Migrate to AssertableJson (Laravel-core)

claudiodekker:
This PR deprecated the assertion library introduced in #220.

Because we added Inertia's assertions to Laravel's core in https://github.com/laravel/framework/pull/36454, it no longer makes sense to maintain two separate versions of this. While it is true that Inertia pioneerd this assertion library, it has since matured as part of the Laravel framework.

In addition to reducing the maintenance effort for the Inertia team, the Laravel variant (by now) also has more features that users might want, [it's properly documented](https://laravel.com/docs/8.x/http-tests#fluent-json-testing), and has the benefit of automatically inheriting assertion macros that are registered by installed third-party packages.

# Differences
Do not worry, **the Laravel variant is nearly identical**. Here's all the differences: 
- The parameter type is different
- The `misses` / `missesAll` methods are named `missing` and `missingAll` instead
- Inertia-specific methods (`component`, `url`, `version`) are **only available on the top-level** when using the new `AssertableInertia` class, or when using `$response->assertInertia(function ($inertia) {`
- The Laravel / new variant has additional methods: [`whereType`](https://github.com/laravel/framework/commit/9981c774a20c9542bf6ed2482a6fe5e1eddc2d30#diff-d301e5322119ba823e76aecc3b468ff94d861dd303d53e30c436c6d2a1977ffc), [`whereContains`](https://github.com/laravel/framework/commit/2d2d108a21b21a149c797cb3995c3a25ac9b4be4#diff-d301e5322119ba823e76aecc3b468ff94d861dd303d53e30c436c6d2a1977ffc), [`each`](https://github.com/laravel/framework/commit/389cd2b9a4075750661d37a1691ecb074c7cfdae#diff-d301e5322119ba823e76aecc3b468ff94d861dd303d53e30c436c6d2a1977ffc) and the ability to do [shorthand-scoping without counting](https://github.com/laravel/framework/commit/fea2e5416f8e5f7fc09b2724d7a6cf4256dbef95#diff-d301e5322119ba823e76aecc3b468ff94d861dd303d53e30c436c6d2a1977ffc)

# Deprecating `Assert`
Right now (Dec 26th, 2021) the `Assert` version is still available, and the `AssertableInertia` is only automatically used on Laravel 8.32 or newer, but a deprecation notice/error will be shown during tests. 

In addition, strict parameter type hints **will crash** in existing tests. To solve this quickly, you can alias your import:

```php
// Before
use Inertia\Testing\Assert;

// After
use Inertia\Testing\AssertableInertia as Assert; 
```

### DIFF
--- a/src/Testing/Assert.php
+++ b/src/Testing/Assert.php
@@ -38,7 +38,7 @@ class Assert implements Arrayable
         echo "\033[0;31m - If you are seeing this error while using \$response->assertInertia(...), please upgrade to Laravel 8.32.0 or higher.\033[0m\n";
         echo "\033[0;31m - If you are using the 'Assert' class directly, please adapt your tests to use the 'AssertableInertia' class instead.\033[0m\n";
         echo "\033[0;31mFor more information and questions, please see https://github.com/inertiajs/inertia-laravel/pull/338 \033[0m\n\n";
-        @trigger_error("Inertia's built-in 'Assert' library will be removed in a future version of inertia-laravel", \E_USER_DEPRECATED);
+        @trigger_error("Inertia's built-in 'Assert' library will be removed in a future version of inertia-laravel: https://github.com/inertiajs/inertia-laravel/pull/338", \E_USER_DEPRECATED);
 
         $this->path = $path;
 
--- a/src/Testing/Assert.php
+++ b/src/Testing/Assert.php
@@ -34,6 +34,12 @@ class Assert implements Arrayable
 
     protected function __construct(string $component, array $props, string $url, string $version = null, string $path = null)
     {
+        echo "\033[0;31mInertia's built-in 'Assert' library will be removed in a future version of inertia-laravel:\033[0m\n";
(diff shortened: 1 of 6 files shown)


RobertBoes:
Does this mean `AssertableInertia` and `$this->assertInertia()` will also be depricated somewhere in the future? Because this also changes how we're supposed to write tests, previously we could use this:
```php
$this->get('/home')
  ->assertOk()
  ->assertInertia(
    fn (Assert $page) => $page
      ->component('Home')
  );
```
Note the "normal" request, not explicitly requesting JSON.

With this change we now have to instruct to use JSON and pass the `X-Inertia` header, because otherwise no JSON would be returned and these assertions won't work. Previously Inertia took this into account, because all props are available in the Inertia response as "sort of JSON", without Inertia assertions it would look like this:
```php
$this->getJson('/home', ['X-Inertia' => true])
  ->assertOk()
  ->assertJson(
    fn (AssertableJson $page) => $page
      ->component('Home')
  );
```

PR #344, opened 2022-01-07
Update CHANGELOG.md

francislavoie:
It's not PHP 5.4 support that was removed, it was Laravel 5.4.


claudiodekker:
Yikes, good catch! Thanks!

PR #347, opened 2022-01-12
Add Laravel 9 Support

claudiodekker:


wiefunkdai (commented):
Fix double base path in url when without serve

PR #350, opened 2022-01-14
Redirect back when an Inertia response is empty (customizable)

claudiodekker:
This PR automatically converts empty Inertia controller/action responses to Inertia redirects back. 

This prevents the need to manually specify the `redirect()->back();` at the bottom of all of your actions, keeping your code a little bit simpler. Do note that this doesn't mean you _have to_ stop redirecting back manually if you prefer to have a consistent return value, but it does provide the option to omit it for those who'd rather not do it everywhere.

This behavior is customizable by customizing the `onEmptyResponse` method that's now part of the Middleware.
You can, for example, render an error page of your own using `Inertia::render()`, or throw an exception.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -4,6 +4,7 @@ namespace Inertia;
 
 use Closure;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Redirect;
 use Symfony\Component\HttpFoundation\Response;
 
 class Middleware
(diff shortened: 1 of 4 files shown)


innocenzi:
This looks nice and would certainly improve the DX.

As discussed on Discord: I suggest adding a property check (or something else) to allow for disabling the feature, to account for edge cases or other instances where this could be not needed.

claudiodekker:
Alright, going to tweak on this a little.

claudiodekker:
Updated. Also updated the original description.

innocenzi:
That's great, good changes 👍

PR #353, opened 2022-01-17
Fix incorrect type hint in render method

### LINKED ISSUE #352: Passing Arrayable object as props no longer working
A recent commit added an `array` type hint to the [ResponseFactory::render()](https://github.com/inertiajs/inertia-laravel/blob/7f06e9b6ffb90f8eab1785ea5bdfd87e2e0df334/src/ResponseFactory.php#L88) method. This broke several of our projects badly because we had been passing in objects with `Arrayable` interface.

For backward compatibility type hint should be `array|Arrayable`.

rhysemmerson:
Fixes #352 

Inertia has always supported passing `Arrayable` objects for props. The type hint added recently broke this feature. ~~This PR adds `Arrayable` to the typehint.~~ I've removed the type hint because older php versions don't support union types

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -85,7 +85,7 @@ class ResponseFactory
         return new LazyProp($callback);
     }
 
-    /** 
+    /**
      * @param  string  $component
      * @param  array|Arrayable  $props
      */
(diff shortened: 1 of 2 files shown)


claudiodekker:
Thanks!

PR #357, opened 2022-01-22
Add ifInertiaSSR blade directive

HassanZahirnia:
This PR adds a `ifInertiaSSR` blade directive which provides a way to check if the app is running in SSR.
```blade
@ifInertiaSSR
   My app is running in SSR mode! 🚀
@else
   My app is not running in SSR 😥
@endif
```

This also helps new users who're trying to use SSR for the first time and struggle to figure out if they're using SSR or not.

### DIFF
--- a/src/Directive.php
+++ b/src/Directive.php
@@ -64,7 +64,7 @@ class Directive
 
             if (isset($__inertiaSsr) && $__inertiaSsr instanceof \Inertia\Ssr\Response ):
         ?>';
-        
+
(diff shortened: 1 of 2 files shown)


claudiodekker:
Hey @HassanZahirnia,

I really don't think this is necessary, as you can simply rightclick -> view source and see based on the JS-rendered HTML being present there that SSR is active. If SSR isn't active, you'll see an empty root <div> element.

PR #358, opened 2022-01-23
Add Realtime Validation feature

HassanZahirnia:
Based on this [PR](https://github.com/inertiajs/inertia/pull/1061). This PR introduces the backend support for the client-side feature.

You simply pass your validation rules to the `Inertia::validate()` function like so:
```php
// Your controller
public function store()
{
   Inertia::validate([
      'first_name' => ['required', 'min:3'],
      'last_name' => ['required', 'max:3'],
   ], [
     'required' => 'The :attribute field is required.', // You can pass custom validation message
   ]);

   // Execution doesn't reach here if validation fails.
   // Also if you pass call `validate` method with no arguments, nothing will happen

   // Further controller logic
}
```

- [x] Re-write the current implementation with a more optimal solution ( Thanks to @innocenzi )
- [x] Improve support/check for validation rules
- [x] Remove the usage of `sometimes` rule and use a different approach
- [x] Add support for custom validation messages
- [ ] Add support for input values that depend on other fields such as `{field}_confirmation`, `required_if`
- [ ] Add support for FormRequests

The current implementation may not be very optimal and will probably be tuned in future commits, so if you have any ideas on how to improve this, don't hesitate to share them!

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -149,9 +149,8 @@ class ResponseFactory
                 );
                 $validator->validate();
             }
-            // We throw this exception to stop the controller from continuing
-            // Since at this point, we know the request has a 'X-Inertia-Validate' header
-            // And because we're doing real-time validation, we don't want the controller to continue
+            // We throw this exception because at this point, we know the request has a 'X-Inertia-Validate' header
+            // And since we're doing a real-time validation, we don't want the controller to continue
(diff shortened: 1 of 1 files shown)


reinink:
Closing based on comment here: https://github.com/inertiajs/inertia/pull/1061#issuecomment-1168107157

Thanks either way ❤️

PR #368, opened 2022-02-14
Fix regression introduced in #337

d8vjork:
We cannot upgrade Inertia from 4.5.0 to 5.x as this introduces a new issue on our software.

### Issue

image

### Scenario

This is a piece of what we've in our `app/Providers/RouteServiceProvider.php`

```php
   /**
     * These namespaces is applied to your controller routes.
     *
     * In addition, it is set as the URL generator's root namespace.
     *
     * @var array
     */
    protected $namespaces = [
        'web' => 'App\Http\Controllers',
        'api' => 'App\Http\Controllers\Api',
    ];

    /**
     * Define your route model bindings, pattern filters, etc.
     *
     * @return void
     */
    public function boot()
    {
        $this->configureRateLimiting();
        $this->configureRoutePatters();

        $this->routes(function () {
            Route::middleware('web')
                ->namespace($this->namespaces['web'])
                ->group(base_path('routes/web.php'));

            Route::prefix('api')
                ->middleware('api')
                ->namespace($this->namespaces['api'])
                ->group(base_path('routes/api.php'));
        });
    }
```

So having this namespace customised for our `web.php` routes file generates this issue of Laravel not detecting that the Inertia **route controller from the macro its actually relative and not absolute as it was before**.

This could also happen with anything that modifies the namespace like route groups, etc. So I think is better having this as a string.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -77,7 +77,7 @@ class ServiceProvider extends BaseServiceProvider
     protected function registerRouterMacro(): void
     {
         Router::macro('inertia', function ($uri, $component, $props = []) {
-            return $this->match(['GET', 'HEAD'], $uri, Controller::class)
+            return $this->match(['GET', 'HEAD'], $uri, '\Inertia\Controller')
                 ->defaults('component', $component)
                 ->defaults('props', $props);
         });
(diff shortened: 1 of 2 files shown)


reinink:
Hey thanks so much for this PR. I just bumped into this myself tonight, so I fixed it (e026fe8e05bba4267d0810727f872405bc225f29), but now I'm realizing that you already submitted a fix 🤦‍♂️

Thanks so much either way for this! I hope to get this released soon.

d8vjork:
@reinink Lovely! I was waiting using a forked for my company project, now we can go again to this

PR #372, opened 2022-02-22
Repeat BaseUrl in navigation

asmodai1985:
concat $request->getBaseUrl() and $request->getRequestUri()  is unreasonable。


reinink:
Hey this was added for an edge case with proxies: #333

I think it's best to leave it 👍

PR #376, opened 2022-03-17
Fix redirect when PUT/PATCH/DELETE route called while unauthenticated

RobertBoes:
The Laravel auth middleware is executed before the `HandleInertiaRequests` middleware, this means when an unauthenticated exception happens Inertia does not interfere. This is an issue when, for example, a DELETE request is made while the user is unauthenticated, since Inertia does not change the statuscode of the redirect from 302 to 303, resulting in a `MethodNotAllowedHttpException`. This PR fixes the issue by letting Inertia decide how the `AuthenticationException` is rendered.

Steps to reproduce:
- Login
- In devtools delete your cookies
- Click on an Inertia DELETE link, the route has to have the `auth` middleware
- The DELETE request is made, then a 302 is returned to `/login`, which is made with as a DELETE request

I found it the easiest to reproduce with a new Laravel project, installed Jetstream, logged in and went to the profile page. There I deleted the cookies and clicked on "Delete account", filled in a random password and then the exception shows

This discussion inertiajs/inertia#1130 explains it in a bit more detail. I thought Inertia would convert the redirect appropriately, but the Inertia middleware is actually never executed

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -2,6 +2,8 @@
 
 namespace Inertia;
 
+use Illuminate\Auth\AuthenticationException;
+use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Foundation\Testing\TestResponse as LegacyTestResponse;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Router;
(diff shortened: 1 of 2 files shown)


RobertBoes:
Unfortunately this approach doesn't work for Laravel 6 and 7, because `renderable` was introduced in Laravel 8. I'm going to take a look at a different approach for this, but so far I wasn't able to find one

alanbernale:
Any solution for this problem? Thank you for your answer.

PR #380, opened 2022-03-22
skip validating errors if session isn't loaded

timrspratt:
This change prevents an error occurring when the middleware is used in a group that doesn't load run `StartSession`


reinink:
Good call, thanks for the PR 👍

Issue #370, opened 2022-02-17
[IDEA] Custom classes on the root component


c-fitzmaurice:
```
@inertia('custom-app-id', 'h-full bg-white')
```

Rendering:

```html
<div id="'.$id.'" class="'.$classes.'" data-page="{{ json_encode($page) }}"></div>
```

This could come in handy when you need to add classes to the main div without having to overwrite it or nest the app inside further components.

RobertBoes:
I'm not sure if Inertia should include this, because of different ways frontend frameworks might render/mount. As an example, Vue 2 replaces the mount point with the mounted component:
```html
<body>
  <div id="app"></div>
</body>

<script>
  const vm = new Vue({
    template:`<h1>Vue 2 Demo</h1>`
  })
  vm.$mount("#app")
</script>

<!-- Output -->
<body>
  <h1>Vue 2 Demo</h1>
</body>
```
Vue 3 does mount inside the mounting point, which would result in the following output:
```html
<body>
    <div id="app"></div>
</body>
<script>
  const app = Vue.createApp({
    template:`<h1>Vue 3 Demo</h1>`
  })
  app.mount("#app")
</script>

<!-- Output -->
<body>
  <div id="app">
    <h1>Vue 3 Demo</h1>
  </div>
</body>
```
These differences between frameworks means it won't be a reliable solution and will only work for certain frameworks

nurdism:
> I'm not sure if Inertia should include this, because of different ways frontend frameworks might render/mount. As an example, Vue 2 replaces the mount point with the mounted component, which means the added classes won't have any effect:

I don't see why that would be an issue. If it doesn't work on/for a given framework then that shouldn't stop this from being included. Simply don't use it if your framework doesn't support it. This would be a very helpful feature as there is no way to accomplish this w/o modifying the package.

riabiy-denis:
Meanwhile I consider this an east-fix
https://stackoverflow.com/a/72671986/6252398

mreduar:
Another idea would also be to add it here.
```
createInertiaApp({
  class: 'h-full bg-white',
  // ...
})
```

lucaSabato:
Any update?

Justinborzi:
@lucaSabato if you want to fix this yourself @nurdism has a PR for it https://github.com/inertiajs/inertia-laravel/pull/382

PR #382, opened 2022-03-25
Ability to add class to @inertia directive

nurdism:
This implements #370

### DIFF
--- a/src/Directive.php
+++ b/src/Directive.php
@@ -12,7 +12,9 @@ class Directive
      */
     public static function compile($expression = ''): string
     {
-        $id = trim(trim($expression), "\'\"") ?: 'app';
+        $args = explode(',', $expression);
+        $id = !empty($args[0]) ?  trim(trim($args[0]), "\'\"") : 'app';
+        $classes = isset($args[1]) ? trim(trim($args[1]), "\'\"") : '';
 
         $template = '<?php
             if (!isset($__inertiaSsr)) {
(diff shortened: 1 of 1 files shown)


nurdism:
@reinink is there something about this PR that is incorrect?

nurdism:
Ahh I see, what about this approach? 

https://github.com/nurdism/inertia-laravel/blob/issue-370/src/Directive.php
https://github.com/nurdism/inertia/blob/issue-370-patch-1/packages/inertia-vue3/src/createInertiaApp.js

I'm not very familiar with react ssr, not sure how to pass the `classes` string to the `createElement` function.

nurdism:
I just realized this doesn't work if you use the `compileHead` function, but maybe this could be added to the inertia config and do the same thing.

nurdism:
I've added a config option 'app' that allows you to pass 'classes' string from it, the `@inertia` directive should function normally w/o any added prams, it defaults to input, then config and then the standard 'app' and reads the classes string directly from the config.

With the ssr request it modifies the `$page` and adds `'classes'` to the request, it pulls the classes string from the config. That can then be picked up and passed the `createElement` or `h` function.

I noticed that id also defaults to app on the SSR side of things, I could also make it pull the id from the config the same way.

Justinborzi:
@reinink is this PR ever getting merged or is this something I should just add to my project manually instead?

Issue #74, opened 2019-09-26
[Idea] View/Component Composers


rojtjo:
One thing which would be helpful in my specific application would be having a feature such as the view composers in Laravel. Currently I'm building a simple solution to do this in my own project, but I'm wondering if there would be any interest of putting this in the core?

I'd be happy to create a pull request for this.

Let me know what you think!

reinink:
What specifically do you mean—preassigning data to specific page components, similar to sharing (global) data?

rojtjo:
Exactly that!

Currently I have a group of pages which require the same dataset. Since it's quite a lot of data I don't want to include this on every response using `share()` but bind it to a specific set of components.

Ideally I'd even have access to the props originally passed to the `render()` method.

thoresuenert:
@reinink the idea is:

replace a Laravel view with a inertia page and use everything from the core the way it is?

```
 View::composer(
            'Pages/Profile/*', 'App\Http\View\Composers\ProfileComposer'
        );
```
Applies the Composer to all Inertia Components.

reinink:
One option is to use the existing `Inertia::share()` functionality, and simply pass the current `Inertia\Response` as an argument.

```php
Inertia::share('mydata', function (Inertia\Response $response) {
    if ($response->component === 'Users\*') {
        return [
            // some data
        ];
    }
});
```

Only issue there is this will always return the `mydata` key.

rojtjo:
I've created a small demo version of my idea here [rojtjo/inertia-laravel](https://github.com/rojtjo/inertia-laravel). Currently it only works using Closure based composers and without wildcard components.

Usage would be something like this:
```php
Inertia::composer('Foo/Bar', function (Response $response) {
    $response->with('foo', 'bar');
});

// Or bind it to multiple components
Inertia::composer(['Foo/Bar', 'Foo/Baz'], function (Response $response) {
    $response->with('foo', 'bar');
});
```

tanshiqi:
@reinink , I can not get the correct result:
```
Inertia::share([
            'something' => function (\Inertia\Response $response) {
                return $response->component;
            },
]);
```
It returns: 'Unresolvable dependency resolving [Parameter #0 [ <required> $component ]] in class Inertia\Response'

reinink:
Just adding a discussion I had on Twitter, related to this:

Brad:

[...]

PR #384, opened 2022-04-02
New Feature: Props Composer inspired by Laravel's View Composer.

ambengers:
Hello Team!

This PR adds a Props Composer feature, which is inspired by View Composers in Laravel. I really think this would be a very good addition to Inertia.js.

## How does this work

Just like in Laravel, you can use `Inertia::composer()` method in any service provider to set composers for specific components. You can use a composer class or a callback function, like so...

```
use Inertia;
use Inertia\ResponseFactory;

class AppServiceProvider extends ServiceProvider
{
    public function boot()
    {
        // Class-based approach..
        Inertia::composer('User/Profile', UserComposer::class);
        
        // Callback-based approach..
        Inertia::composer('User/Profile', function (ResponseFactory $inertia) {
            $inertia->with([
                'post' => [
                    'subject' => 'Hello World!', 
                    'description' => 'This is a description.'
                 ]
            ]);
        });
    }
}
```

Then your composer class must contain a `compose()` method that will accept an instance of `Inertia\ResponseFactory` class as the first argument. Within this method you can call `$inertia->with();` to set the composed props. This method can accept an array or key-value arguments.

```
use Inertia\ResponseFactory;

class UserComposer
{
    public function compose(ResponseFactory $inertia)
    {
        $inertia->with('list',  ['foo' => 'bar', 'baz' => 'buzz']);
    }
}
```

Of course, you can also use the same composer for multiple components like so...

```
Inertia::composer(['User', 'User/Edit'], UserComposer::class);
```

### DIFF
--- a/src/Composer.php
+++ b/src/Composer.php
@@ -13,7 +13,7 @@ class Composer
     protected $composers = [];
 
     /**
-     * @param ResponseFactory $factory
+     * @param  ResponseFactory  $factory
      */
     public function __construct(ResponseFactory $factory)
     {
@@ -21,8 +21,8 @@ class Composer
     }
 
     /**
-     * @param string $component
-     * @param Closure|mixed $composer
+     * @param  string  $component
+     * @param  Closure|mixed  $composer
      */
     public function set($component, $composer)
     {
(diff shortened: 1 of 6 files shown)


ambengers:
This PR implements #74 .

PR #377, opened 2022-03-18
Update Inertia.php

SuryaWebfox:
Remove incorrect type hint: https://github.com/inertiajs/inertia-laravel/pull/353

PR #387, opened 2022-04-20
Fix `render` method docblock on facade

hailwood:
The render method can take an array or an Arrayable.
This inconsistency was introduced when #353 was merged.

If we're unhappy with this due to union types, I would suggest we look at merging #377


reinink:
@hailwood Hey thanks!

PR #398, opened 2022-05-22
Set `Vary: Accept` header for HTML responses

reinink:
Currently we only set the `Vary: Accept` header for Inertia (JSON) responses, however we also need to set this header for standard full page HTML responses as well, otherwise it causes issues in Safari.

These issues may include seeing JSON on a page instead of rendered HTML, or getting an Inertia modal (only meant for showing errors) for proper Inertia pages.

This mostly worked previously because we set the `Vary: Accept` header on all Inertia (JSON) responses, and I guess most browsers were able to figure it out based on that, but Safari is a little more particular and requires it to be on both the Inertia (JSON) responses as well as the standard full page HTML responses.

This PR actually updates ever single response to include this header — even those that are not Inertia endpoints at all. This feels maybe a little heavy handed, but from my testing this seems to be required, as it appears to even be required on redirect responses that end up on Inertia endpoints.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -87,6 +87,8 @@ class Middleware
 
         $response = $next($request);
 
+        $response->header('Vary', 'Accept');
+
         if (! $request->header('X-Inertia')) {
             return $response;
         }
--- a/src/Response.php
+++ b/src/Response.php
@@ -104,10 +104,7 @@ class Response implements Responsable
         ];
 
         if ($request->header('X-Inertia')) {
-            return new JsonResponse($page, 200, [
-                'Vary' => 'Accept',
-                'X-Inertia' => 'true',
-            ]);
+            return new JsonResponse($page, 200, ['X-Inertia' => 'true']);
         }
 
         return ResponseFactory::view($this->rootView, $this->viewData + ['page' => $page]);


crynobone:
Hi @reinink 

I believe adding the middleware this way is too aggressive where it doesn't check for the type of response. This has been demonstrated in the following issue where `StreamedResponse` and `BinaryFileResponse` where both methods doesn't have `header()` setter method and there is absolutely no reason for them to have `Vary: Accept` header value. 

I would assume that it would be best to check if the return value is `Content-Type: text/html` and have the` header()` setter method.

Regression issues:

* #400 
* #401 
* This also breaks any Laravel Nova application using Export As CSV (which use `StreamedResponse`)

image

crynobone:
> Should probably just be for requests that want HTML. https://github.com/inertiajs/inertia-laravel/pull/402#issuecomment-1136493369

After some time on my side, it still might be tricky to just to verify normal HTML vs download/streamed response. 

### Normal HTML Request

Request has `Accept: text/html` and Response has `Content-Type: plain/text; charset: utf-8`

image

### Nova's Streamed Response Request

Request has `Accept: application/json` and Response has `Content-Type: plain/text; charset: utf-8`

image

reinink:
Yeah, it honestly may not matter if we set the `Accept` header all the time…we just need to use a header method that exists on all the responses 🤦🏻‍♂️

PR #399, opened 2022-05-24
Add asset versioning for default build location using vite

WouterSioen:
When mix is not used for asset compilation, chances are high that vite is used instead.
That's why I think it makes sense to add the default location that vite builds to as a
fallback for when now mix-manifest is available.

https://laravel-vite.dev/guide/essentials/building-for-production.html#build-path contains
some information about the default production build using laravel-vite.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -36,6 +36,10 @@ class Middleware
             return md5_file($manifest);
         }
 
+        if (file_exists($manifest = public_path('build/manifest.json'))) {
+            return md5_file($manifest);
+        }
+
         return null;
     }
 


reinink:
Hey thanks for this. Is the `build/manifest.json` path specific to the Laravel Vite package, or is this just a Vite default when running a build?

WouterSioen:
This is mostly specific to the Laravel Vite package. The default build folder for vite is the `dist` folder. This makes sense in an SPA where you just deploy the frontend code but not really in a Laravel + Inertia app where you deploy your frontend code inside your public folder to make it accessible from the browser.

jessarcher:
@reinink This would also work for the official Laravel Vite integration as it uses the same path.

reinink:
Great, thanks @WouterSioen and @jessarcher — that makes sense to me 👍

bcalik:
btw why not use something faster than `md5`, like a non-cryptographic hashing algorithm like `crc`, which is also used for detecting unintentional errors in data.

Issue #401, opened 2022-05-24
Call to undefined method Symfony\Component\HttpFoundation\BinaryFileResponse::header()


mberneis:
When I upgraded my application to the latest and greatest (php8.1, latest Laravel) I encountered this error:

```
 Error: Call to undefined method Symfony\Component\HttpFoundation\BinaryFileResponse::header() in .../vendor/inertiajs/inertia-laravel/src/Middleware.php:90
```

Looks like it can be fixed by changing line 90 in `src/Middleware.php`
``` 
$response->header('Vary', 'Accept');
```
 to 
```
$response->headers->set('Vary', 'Accept');
```
according to https://laracasts.com/discuss/channels/laravel/call-to-undefined-method-binaryfileresponseheader-1

reinink:
Thanks for reporting this — this has been fixed and published: https://github.com/inertiajs/inertia-laravel/releases/tag/v0.6.2

PR #402, opened 2022-05-24
Changed header settings in Middleware to support latest Symphony library

mberneis:
Fix for Issue #401

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -87,7 +87,7 @@ class Middleware
 
         $response = $next($request);
 
-        $response->header('Vary', 'Accept');
+        $response->headers->set('Vary', 'Accept');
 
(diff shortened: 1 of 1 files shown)


crynobone:
@mberneis it's weird to have the need to add that header for file download response. The same goes to stream response.

@reinink This clearly breaks existing applications including Nova.

image

reinink:
Yup, shoot, didn’t think the header API would be different between response types. That’s surprising. And you’re right, we don’t technically need to add the accept header to everything. Should probably just be for requests that want HTML. Out at my daughter‘s baseball game at the moment, but I’ll try get this fixed tonight.

mberneis:
@reinink - closing my pull request as your fix makes much more sense.

PR #403, opened 2022-05-24
Check middleware Response instance before trying to modify for Inertia

### LINKED ISSUE #400: Undefined method Symfony\Component\HttpFoundation\StreamedResponse::header()
Hi,

In one of my projects I'm making use of a streamed response, so `return response()->stream()`. Since the upgrade to 0.6.1 I'm seeing the following error:

`Error: Call to undefined method Symfony\Component\HttpFoundation\StreamedResponse::header() in file /var/www/<projectname>/vendor/inertiajs/inertia-laravel/src/Middleware.php on line 90
`

Probably related to changes made in #398?

### LINKED ISSUE #401: Call to undefined method Symfony\Component\HttpFoundation\BinaryFileResponse::header()

crynobone:
* fixes #400 
* fixes #401

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -4,9 +4,9 @@ namespace Inertia;
 
 use Closure;
 use Illuminate\Http\JsonResponse as LaravelJsonResponse;
-use Illuminate\Http\Response as LaravelResponse;
 use Illuminate\Http\RedirectResponse as LaravelRedirectResponse;
 use Illuminate\Http\Request;
+use Illuminate\Http\Response as LaravelResponse;
 use Illuminate\Support\Facades\Redirect;
 use Symfony\Component\HttpFoundation\Response;
 
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -5,7 +5,10 @@ namespace Inertia;
 use Closure;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Redirect;
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\StreamedResponse;
 
 class Middleware
 {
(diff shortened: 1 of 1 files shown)


crynobone:
Alternatively we could just set the header from Response https://github.com/inertiajs/inertia-laravel/blob/a232be0c66de6552314e32effaf78b3dd16a20b0/src/Response.php#L110

reinink:
> Alternatively we could just set the header from Response

I tried this originally, but if there is a redirect between two Inertia endpoints, even that redirect response needs the `Vary` header set, otherwise it doesn't work — at least from my testing in Safari, where this issue exists.

crynobone:
> I tried this originally, but if there is a redirect between two Inertia endpoints, even that redirect response needs the Vary header set, otherwise it doesn't work — at least from my testing in Safari, where this issue exists.

Check the code to explicitly check for `Illuminate\Http\Response`, `Illuminate\Http\JsonResponse` or `Illuminate\Http\RedirectResponse`. This should cover most of the use-case and allows `Accept: Vary` to be skipped from any custom Symfony based Response.

reinink:
@crynobone Can you think of any practical reason why including `Accept: Vary` on those other responses is bad? Because as far as I can tell it just tells the browser to be more careful about how it caches the response.

reinink:
> I don't necessarily feel that it's bad. Just that it changes the header on responses unrelated to Inertia scope (initial HTML page, redirects, and JSON responses). If those response needs to have the header it should be explicitly defined by the related code.

Yeah, I don't disagree, but again, from my testing, it doesn't work unless it's on at least all Inertia endpoints (HTML and JSON), *plus* any 301 redirects that end up on an Inertia endpoint.

Further, I'm starting to think that `Vary: Accept` is actually the wrong header to begin with. Like, it's been working to this point, but I think mostly by chance. I think the header we should be using is `Vary: X-Inertia`.

I'm working on a PR that implements all of that — which I think is the better fix this what's in this PR.

PR #410, opened 2022-06-02
Make automatic redirect opt-in instead of opt-out

RobertBoes:
This PR makes the automatic redirect back from an empty response opt-in instead of the current opt-out. 

The reasoning behind this: I just created a fresh Inertia install and started building with TDD. After writing a test I ran it and immediately I was confused, one of the very first failures was an unexpected status code. Instead of a 200 I received a 302. I searched for quite some time, thought it might've been an incorrect middleware or something else really funky going on. Looked through Inertia's [release notes](https://inertiajs.com/releases) and didn't see anything obvious that could cause it (I later noticed the releases on the website aren't maintained). After digging for a while I noticed #350 introduced a redirect back when an empty response is returned. Enabling this by default is extremely confusing imo, since a default Laravel app will never do this, but instead it creates an empty 200 response.


reinink:
Hey @RobertBoes! 👋 

So while I get your point, and this was a concern I had at first when adding this feature, I actually think it's fine as it is, since this behavior *only* happens on Inertia requests/responses. If it happened for *all* requests, then I think I'd agree with you.

That said, I think it should definitely be documented better. I am going to add this to my list of things to add to the website.

Hope that makes sense, and thanks either way for the contribution 👍

PR #419, opened 2022-06-22
Preserve props on partial reloads

lepikhinb:
This PR introduces an ability to preserve properties on partial reloads even if they aren't specified on a request.

```php
Inertia::preserveProperties(['alert']);
```

While `only` is great within a context, there could be a global state with properties that need to present on every `get` request. So, it makes sense to control them from a backend to ensure that required props are always present on the response. 

Examples:
- Alert notifications
- Popups (onboarding)
- Auth user data

So, any global state that's independent of a current context.


lepikhinb:
See https://github.com/inertiajs/inertia-laravel/pull/621

Issue #421, opened 2022-06-23
Subdirectory urls broken


andrew1601:
#347 breaks Laravel projects that run in a subdirectory.

src/Response.php was changed at line 128 but nothing in the commit message details why. Symfony don't seem to have changed what these methods do and this now returns the incorrect URL which causes the Inertia client adapter to push an incorrect path to the history. So an Inertia app running at https://example.com/app will have https://example.com/app/app pushed to the history on the initial page render automatically.

The issue is also described here: https://laracasts.com/discuss/channels/laravel/root-directory-is-displayed-twice-in-url-on-load-and-reload

PR on the way

CristianSitov:
I'm watching all these issues: current, #422 & #437 
Hopefully, one of them will be merged soon.

PrasadChinwal:
Any update on this issue?

claudiodekker:
https://github.com/inertiajs/inertia-laravel/pull/446#issuecomment-1920171976

jessarcher:
Hi, this should be addressed by #592. Sorry for the delay!

driesvints:
Just FYI: Laravel has never and never will support sub directories and you should avoid running them as such: https://laravel.com/docs/10.x/installation#directory-configuration

Dotgix-Technologies:
i waw facing issue on suburls then urls wars repeating it self so after hours of debugging i found the issue no neet to change in any ventor file just define 
 <base href="/public">
then it work for me

PR #422, opened 2022-06-23
correct url passed through in responses

andrew1601:
Fix for #421

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -99,7 +99,7 @@ class Response implements Responsable
         $page = [
             'component' => $this->component,
             'props' => $props,
-            'url' => $request->getBaseUrl().$request->getRequestUri(),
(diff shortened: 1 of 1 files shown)


sinceow:
The problem still exists in v0.6.4 @reinink, please merge the commits.

Rubens10010:
This works for me too! I've been using a custom algorithm to bypass this error but this was it. You should merge it ASAP

jessarcher:
Thanks for the PR, and I'm sorry for the delay! Unfortunately, this PR would have re-introduced the bug that https://github.com/inertiajs/inertia-laravel/pull/333 was trying to fix, so I've merged https://github.com/inertiajs/inertia-laravel/pull/592 which should address both use cases.

PR #425, opened 2022-06-29
Log SSR errors in Laravel

alexryall:
Currently the only place SSR errors are logged is in the output of the Node process which is not particularly useful for production systems. Several times we have managed to make code changes live and accidentally break the SSR without knowing.

This change will accompany another PR on [intertia](https://github.com/inertiajs/server) so that SSR errors are included in the Node server response and logged in Laravel via whichever adapters have been setup in Laravel.

The isset is required to make the change backwards compatible with the node server when it doesn't respond with a body.

This is an example of an SSR error logged using the https://github.com/TheoKouzelis/laravel-airbrake adapter
image


vasilGerginski (commented):
i like that. I am having trouble with SSR as well.

driesvints:

PR #432, opened 2022-07-19
Fix unexpected JSON resource serialization

TheDoctor0:
I was in the process of porting the project to Inertia and I encountered a very unexpected issue with the JSON resources serialization. Some had a different structure in comparison to the one returned from API routes - `data` wrapper was present.

I tinkered with the source code and discovered that the issue lies in how resources are being serialized in the response https://github.com/inertiajs/inertia-laravel/blob/master/src/Response.php#L136.

I was a little surprised to find out that the `FakeResource` used for tests has a `$wrap` property set to `null` to mitigate this issue but that has not been documented anywhere. Without this property, the test for response with JSON resource fails.

To fix this unexpected behavior I changed the logic for resources serialization that will make sure to not use the `data` wrapper.


driesvints:
Seems something went wrong with branching here

TheDoctor0:
@driesvints I only synchronized the fork. I will open a fresh PR.

PR #433, opened 2022-07-22
add dumpInertia() testing method

JulienCauvin:
This pull request intent is to add a new `dumpInertia(string $key = null)` method to TestResponseMacros file. Existing testing methods like `dump()`, `dumpSession()`, `dumpHeaders()` are very convenient in Laravel testing.

If parameter is not set, then all of the Inertia data will be printed (component, props, url, version, ...).
If parameter is set, then props property content will be printed (accepting dot-notation thanks to `Arr:get()`)


driesvints:

PR #435, opened 2022-08-07
feat: support nested and wildcard `only` props

Tofandel:
This PR adds support for lazy props at any level and retrieving partial data at any depth

Eg: `only: ['modal.props.aLazyProp']` or `only: ['allNestedLazyProps.*']`

There is support for 2 type of wildcards
```
modal.* => will load all nested props under "modal" and lazy props that are direct children of modal
This wildcard can be continued with the dot notation. Eg:
modal.*.props => will only load all props (lazy or not) named "props" which are 2 level deep within modal

modal.** => will load all lazy and non lazy props under modal, no matter how deep
```

It's a necessary feature if we want to implement modals and such and very useful for improving performances of very complex pages with a lot of data

Goes by par with https://github.com/inertiajs/inertia/pull/1247


innocenzi:
> It's a necessary feature if we want to implement modals and such

Why is that?

Can the implementation be simplified using Laravel's `Arr::dot` and `Arr::undot`?

Tofandel:
Because if you implement a modal you usually need a lazy prop "modal" under which you provide the modal data, 
But this modal might have itself lazy data that you want to explicitely request for, so you need to do this

```php
"modal" => Inertia::lazy(fn () => [
  "lazyProp" => Inertia::lazy(fn () => "abc"),
]
```

Which is impossible to do with the current implementation


It couldn't be simplified with Arr:dot and Arr:undot for one simple reason, you'd have to evaluate lazy props before if you want to flatten it in a nested manner and run that on it, it would also perform 10 times worse as you need to perform the flattening and unflattening on the props recursively which could very well have a 10 000 elements especially if you share models with relations instead of the "only" value which is usually only 1-3 elements, you also run the risk of loosing elements (I don't know if Arr::dot preserves empty arrays) but it's overall not a good idea, you really want to work with a tree for this kind of value visiting

Tofandel:
Gave it a try and yes, the issue is that `Arr::undot` will not handle correctly

`only: ['modal.lazyProp', 'modal']` (If you want to reload both the lazy and the normal props of modal)

Because it either results in `["modal" => []]` or `['modal' => ['lazyProp' => []]]` and there is no way to represent this structure without a tree but I might have a way around that

driesvints:

onlime:
> Heya. We removed our master branch. Feel free to re-attempt this to 1.x!

@Tofandel yes please! This is definitely needed. Greatly appreciating your work towards a glory revival of Inertia!

Waiting for this for quite some time and going to use it in combination with https://github.com/lepikhinb/momentum-modal/pull/18 👏

Tofandel:
@onlime Then please feel free to take over my PR on 1.x, keeping in mind that half of the feature has been shipped already https://github.com/inertiajs/inertia-laravel/pull/620

I have completely switched from inertia to [hybridly](https://hybridly.dev/) (like inertia but much better) due to my complete dissatisfaction with maintenance, lack of progress and disregard for contributors on inertia (I opened about 6 PRs which took several days to make, none of which were ever as much as reviewed and because of the multiple force push it has become impossible to work with and of course I'm only a needle in the haystack of contributors that have the same stories)

I will not spend any more wasted time on inertia after this debacle

PR #437, opened 2022-08-11
fix: subdirectory project link error

chrisAlkemy:
Working on a project I noticed that when I was under a sub directory the links were generated in a strange way.

Example:

http://www.example.com/project1/first-page

I try to navigate to second-page, it generates this:

http://www.example.com/project1/project1/second-page

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -99,7 +99,7 @@ class Response implements Responsable
         $page = [
             'component' => $this->component,
             'props' => $props,
-            'url' => $request->getBaseUrl().$request->getRequestUri(),
+            'url' => $request->fullUrl(),
             'version' => $this->version,
         ];
 


Tofandel:
See also #421

CristianSitov:
I'm watching all these issues: current, #421 & #422 
I spent almost 2 days figuring out how to trick it via nginx/confs but it seems that this is the only/best workaround.
Hopefully, one of them will be merged soon.

vertexportus:
> I'm watching all these issues: current, #421 & #422 I spent almost 2 days figuring out how to trick it via nginx/confs but it seems that this is the only/best workaround. Hopefully, one of them will be merged soon.

yeah well, the earliest fix PR for this is from back in January - so it's taking a while to get merged and a new version published

jessarcher:
Thanks for the PR, and I'm sorry for the delay! I've merged #592 which should address this issue without adding the scheme and host, which may be a breaking change for some.

PR #454, opened 2022-10-03
Prepare support of PHP 8.2

JamesHemery:
Hi !

Just a small PR to prepare for PHP 8.2 support which we are already using in development on some projects.


.github/workflows/tests.yml
```diff
@@ -15,6 +15,10 @@ jobs:
         php: [ 7.2, 7.3, 7.4, '8.0', 8.1 ]
         laravel: [ 6, 7, 8, 9 ]
         stability: ['prefer-lowest', 'prefer-stable']
+        include:
+          - php: 8.2
+            laravel: 9
+            stability: prefer-stable --ignore-platform-req=php+
```
driesvints, line 21: This line isn't needed anymore. PHP 8.2 is stable enough to start requiring other packages to release compatible versions. We've also removed this from the Laravel framework build and have a passing build. In fact, 8.2 can be added to the list of PHP versions a bit above here. It needs to be excluded for Laravel 6, 7 & 8 in the exclude list below.

driesvints (changes requested):

.github/workflows/tests.yml
```diff
@@ -15,6 +15,10 @@ jobs:
         php: [ 7.2, 7.3, 7.4, '8.0', 8.1 ]
```
reinink, line 15: @driesvints like so?

```suggestion
        php: [ 7.2, 7.3, 7.4, '8.0', 8.1, 8.2 ]
```

.github/workflows/tests.yml
reinink: Just this line (line 21)? Or more?

driesvints:
Preparing a new PR because I can't suggest commits in here on parts that I want to change: https://github.com/inertiajs/inertia-laravel/pull/463

PR #456, opened 2022-10-06
Laravel translation

abdosaeedelhassan:
Adding ability to hold current locale json translation with inertia response, later in vue adapter we can also add translation function like next:
`  __(key, replace = {}) {
            var translation = this.$page.props.language[key]
                ? this.$page.props.language[key]
                : key
            Object.keys(replace).forEach(function (key) {
                translation = translation.replace(':' + key, replace[key])
            });
            return translation
        }`
That give us translation like normal in blade component ex:
`{{__("Hi")}}`


driesvints:
No plans for this sorry.

PR #462, opened 2022-10-27
Allow Inertia\Response rebinding in user-land

pyrou:
For some specific use case, I would be able to override the `Inertia\Response` class to add some project-specific helpers, that can't be achievable with `Inertia::shared()`.

Switching from `new Response` to `app()->make(Response::class)` in ResponseFactory would let us (as project user) rebind the Response class in a Service provider :

```php
$this->app->bind(\Inertia\Response::class, CustomInertiaResponse::class);
```

This PR doesn't introduce breaking changes.

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -97,10 +97,10 @@ class ResponseFactory
         }
 
         return app()->make(Response::class, [
-            "component" => $component,
-            "props" => array_merge($this->sharedProps, $props),
-            "rootView" => $this->rootView,
-            "version" => $this->getVersion(),
+            'component' => $component,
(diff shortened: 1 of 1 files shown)


flavio-paqt:
No tests to confirm?

pyrou:
@flavio-paqt There is existing unit tests that cover enough the ResponseFactory. But not executed yet by GitHub Actions because no maintainer has found time to get some review and click "Approve and run" button on this PR in the last 5 months.

But just like on plenty of other PRs, up to 2 yrs old, contributions looks like dead letters.
So the better I can do is to close this PR and stop wasting my time in the future ;)

boring-dragon:
I actually had a requirement to do something like this for a project of mine.

PR #476, opened 2022-12-16
Output json when TestResponse contains an exception

benbjurstrom:
This pull request outputs a console friendly json response when AssertableInertia is passed an Illuminate\Testing\TestResponse containing an exception.

#### Current Behavior
Tests utilizing the AssertableInertia class return all unhandled exceptions with the same 'Not a valid Inertia response' message. This limited information makes troubleshooting a failing test cumbersome. 

To identify the actual problem I typically need to rerun the test with `->dump()` chained on before the call to assertInertia. This workaround is not ideal as 1) it dumps the Ignition response as raw html/css and 2) is not possible to do in a CI pipeline.

For example:
```php
    $this->get(route('posts.index'))->dump()
        ->assertInertia(fn(AssertableInertia $page) => $page
            ->component('Posts/Index')
        );
```

### DIFF
--- a/src/Testing/AssertableInertia.php
+++ b/src/Testing/AssertableInertia.php
@@ -31,6 +31,9 @@ class AssertableInertia extends AssertableJson
             PHPUnit::assertArrayHasKey('url', $page);
             PHPUnit::assertArrayHasKey('version', $page);
         } catch (AssertionFailedError $e) {
+            if ($response->exception) {
+                $response->json();
+            }
             PHPUnit::fail('Not a valid Inertia response.');
         }
 


benbjurstrom:
Ok, I think I found a workable solution. When the assertSuccessful method fails it dumps a console friendly stack trace. Adding that before calling assertInertia should probably be standard practice.

```php
$this->get(route('posts.index'))
    ->assertSuccessful()
    ->assertInertia(fn(AssertableInertia $page) => $page
        ->component('Posts/Index')
    );
```

PR #484, opened 2023-01-12
Improve commands

datlechin:
#### What changes?

- Change namespace `Console` to `Commands`
- Instead return `0`, `1` we return constant `FAILURE`, `SUCCESS` from base extend class

### DIFF
--- a/src/Commands/CreateMiddleware.php
+++ b/src/Commands/CreateMiddleware.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Inertia\Console;
+namespace Inertia\Commands;
 
 use Illuminate\Console\GeneratorCommand;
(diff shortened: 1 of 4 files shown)


rojtjo:
I don't think it's wise to use the `$components` property as it's marked as `@internal`. While I would probably be fine using this in a project, I don't think it's wise to do the same in a library like this.

Also, the fresh Artisan look wasn't introduced until Laravel 9.21, which means this will not be backward compatible with older Laravel versions that are supported by this package.

reinink:
Hey thanks @datlechin and @rojtjo! That's super helpful.

@datlechin would you mind updating this PR to only include the namespace fixes, method return types, and improved responses, but remove the components stuff? 🙏

reinink:
@datlechin Thanks!

PR #492, opened 2023-01-15
Add pcntl to composer suggestions

RobertBoes:
v0.6.6 introduced artisan commands to start the SSR server. Later this was updated (#487), but the code introduced there uses pcntl, which isn't currently part of the requirements. This adds `ext-pcntl` to the requirements. Since the docs now only provide information to start the SSR server through this command I feel like this requirement should be added to Inertia.

One side-effect of this is that's it's impossible to use this feature on systems where the extension isn't available, like Windows.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
     "require": {
         "php": "^7.2|~8.0.0|~8.1.0|~8.2.0",
         "ext-json": "*",
+        "ext-pcntl": "*",
         "laravel/framework": "^6.0|^7.0|^8.74|^9.0|^10.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "phpunit/phpunit": "^8.0|^9.5.8"
     },
     "suggest": {
-        "ext-pcntl": "Required to run Inertia SSR through the inertia:start-ssr artisan command"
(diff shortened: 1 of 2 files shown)


darthsoup:
It might be better to just include it in the suggestions. Not everyone uses SSR.

RobertBoes:
@darthsoup Yeah you're right. I moved it to `suggest` and added a check to the command, that seems like a better solution, thanks 👍

darthsoup:
@rojtjo depends on the distribution and the operating system. pcntl does not work under Windows

rojtjo:
I know it's not available on Windows, but was just wondering if it's even required to start the SSR server.

Looking at the code it's only used to register a stop handler which kills the SSR process, but as far as I know the SSR process will also be automatically killed when the parent process (`artisan inertia:start-ssr`) stops. This would make the pcntl actually optional and should not prevent this command from running.

reinink:
Hey thanks for this @RobertBoes!

I've just tweaked it a little to conditionally use the `pcntl` extension when it's available, and I updated the suggest message to reflect that it's recommended when using the `inertia:start-ssr` command.

I think this should generally be fine, because even using the `pcntl` functions here was being "extra safe", basically just cleaning up any old node processes if things go bad.

PR #446, opened 2022-09-26
fix duplicates app URL in the URL

akramghaleb:
fix duplicates app URL in the URL

Issue #359, opened 2022-01-27
Error in view data / url prop


kimhf:
Currently if the project has the base url `localhost/test` and we navigate to `some/sub/page` the result is a url change to `localhost/test/test/some/sub/page` in the brower by inertiajs. This url no longer matches the route we are currently at.

The issue seems to be that a path as part of the base url is not handled correctly.

This line in \Inertia\Response causes the issue.

`'url' => $request->getBaseUrl().$request->getRequestUri(),`

rhysemmerson:
This looks like a duplicate of #236

kimhf:
I do not believe this is a duplicate of #236, but a solution might solve both issues.

NickSdot:
@claudiodekker seems this was introduced in https://github.com/inertiajs/inertia-laravel/pull/333

If I am not mistaken this anyway shouldn't be Inertia's job. The behavior before was just fine and the issue which lead to the PR should be solvable by e.g. setting `root` in the NGINX config (untested).

kimhf:
It should in my opinion be Inertia's job to handle different server configurations. The problem as far as I can see is that #333 is the wrong fix for the problem. At least in my case where I'm setting `RewriteBase /to/project` in .htaccess

The combination of `$request->getBaseUrl()` and `$request->getRequestUri()` generates a invalid url with that server config. Looking at how `fullUrl()` / `getUri()` builds the url `$this->getBaseUrl()` + `$this->getPathInfo()` + `$queryString` (or `fullUrl()`) seems to be the right approach and should handle more server configurations.

lotsofbytes:
Thanks @kimhf. I can't wait this gets released.

I have just started using inertia a couple days ago. This will solve my problem with a project in subfolder.

juancarlos-ctg:
Hope this gets fixed soon, it fixed my issue as well. ✊

qbuache:
also interested in a fix, at the moment I have to mess with the toResponse function in /vendor/inertiajs/inertia-laravel/src/Response.php

romanmartushev:
I am also interested in this fix! Currently unable to update a site to Laravel 9 because of this issue.

dopesky:
This also fixed my issue. Hope it gets merged soon.

lucraraujo:
Holy sh**! I lost one week of my life because of this bug!
I tested 3 different web server (apache, nginx, IIS) to finally find this bug!

andrew1601:
[...]

Issue #236, opened 2021-03-17
URL used in Response class does not respect HTTP_X_FORWARDED_PREFIX header


lmjhs:
When using a reverse proxy to remove the path prefix from a URL, the url property sent with the inertia response does not respect the HTTP_X_FORWARDED_PREFIX header.  This causes the browser history pushState to be rewritten to the wrong URL.

For example:
- Initial Request: http://example.com/v2/login
- Reverse proxy strips v2 prefix: http://example.com/login
- Intertia response url: /login
- Browser URL rewritten to: http://example.com/login

Changing `'url' => $request->getRequestUri()` to `'url' => $request->fullUrl()` in `Inertia\Response` seems to solve the problem as this will create a full url including the HTTP_X_FORWARDED_PREFIX header.

However there may be a more elegant solution?

herpaderpaldent:
This does not seem to be an inertia issue. Assuming you are using laravel (hence this repo) have you set, have you configured the `ThrustProxies` to trust any proxy? 
maybe this is a good read for you: https://michielkempen.com/blog/generate-https-urls-when-running-laravel-behind-a-proxy

lmjhs:
Thanks @herpaderpaldent for your quick reply.

Sorry yes this is in a Laravel project and yes I have already configured to TrustProxies so URL generation is working correctly.  Redirects work correctly but when I hit a page managed by Inertia it then updates the pushState in the browser to the URL sent in the Response payload.  So the page at first works correctly but if the user refreshes the page it is now on the wrong page.

The issue is that Inertia is not using the same functions as Laravel for generating the URL in the response.  By using the getRequestUri() method it only returns the path and not the fully formed URL using the HTTP_X_FORWARDED_PREFIX header.  My suggestion is that by changing that to use the fullUrl() method it would include it and from some initial testing appears to fix the problem.  However I think that will return the domain as well as the path so not sure if that would cause any other issues?

garagepoort:
Hello,

I seem to have the same problem. I've setup a reverse proxy to redirect to a specific subpath of the laravel application.
This works fine but the inertia response includes the full path in it's response.

Then the js frontend updates the browser history to include the full path.
So for example I have a domain `https://myexample.com/`
Nginx proxies to `https://mylaravelapp.com/app/143716`
Inertia updates the browser history to: `https://myexample.com/app/143716`. In the response I can see that inertia included the full path rather keeping the original requested path.

[...]

PR #499, opened 2023-02-01
Allow URL override for response

FrittenKeeZ:
Umbrella change for:
- #236
- #359
- #360
- #421
- #422
- #437
- #446

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -81,10 +81,11 @@ class Response implements Responsable
      * Create an HTTP response that represents the object.
      *
      * @param \Illuminate\Http\Request $request
+     * @param string|null              $url
      *
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function toResponse($request)
+    public function toResponse($request, ?string $url = null)
     {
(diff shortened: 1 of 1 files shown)


rojtjo:
The thing I dislike about this is that you'd be required to manually call the `toResponse` method from your controller or middleware.

I would probably prefer a solution where you'd be able to just override a method on the `HandleInertiaRequests` middleware to define a resolver. This way there is no need to manually call the `toResponse` method and the framework can just handle the rest.

```php
use Inertia\Middleware;

{
    {
    }
}
```

I've prepared possible solution for this here: https://github.com/inertiajs/inertia-laravel/compare/master...rojtjo:inertia-laravel:feature/url-override

FrittenKeeZ:
@rojtjo I like that method a lot - I initially thought about making a resolver on the response class itself, but it would basically be the same thing, as you had to call `toResponse` as well.

FrittenKeeZ:
@rojtjo would you consider making a PR with your changes?

rojtjo:
Sure, no problem!

FrittenKeeZ:
Closing this in favor of #503 which works as expected

PR #500, opened 2023-02-10
Allow additional headers to be sent with location response

chrisloftus:
Use case: redirecting to an affiliate link - to stop these links being indexed in search engines, we can send the `X-Robots-Tag: nofollow` response header. This PR allows this.

https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag


src/ResponseFactory.php
```diff
@@ -105,14 +105,18 @@ public function render(string $component, $props = []): Response
     /**
      * @param string|RedirectResponse $url
      */
-    public function location($url): \Symfony\Component\HttpFoundation\Response
+    public function location($url, ?array $headers = []): \Symfony\Component\HttpFoundation\Response
```
jbrooksuk, line 108: This doesn't need to be nullable since a default is set:

```suggestion
    public function location($url, array $headers = []): \Symfony\Component\HttpFoundation\Response
```

driesvints:

PR #503, opened 2023-02-10
Allow URL override by defining a URL resolver

rojtjo:
This pull request provides an alternative solution to @FrittenKeeZ's pull request ( #499) by allowing you to configure a callback that resolves the current URL. This way it can be easily configured to anyone's needs by overriding the `urlResolver` method in the `HandlesInertiaRequests` middleware. When this method returns `null` it falls back to the default behavior.

Example:
```php
use Inertia\Middleware;

{
    {
    }
}
```


mahbubahmed:
@rojtjo  This is an elegant solution to fix broken URLs when the application is inside a sub-directory. 

When can we expect to get this solution merged?

brgmn:
Would really like to see this merged! Thanks @rojtjo for your great work! I think this could also solve running Laravel Nova 4.x in a subdirectory (based on inertia, broken at the moment).

ngekoding:
Any update for this?

Will the PR merged or any other solutions?

PR #531, opened 2023-07-23
Add Defer Prop

adrum:
I felt inspired by Livewire 3 Lazy feature which automatically fetched the data on page load. I think there's a ton of value in Inertia's current Lazy prop feature, but I think there are also scenarios where adding a Defer prop feature which automatically loads Lazy data on page load.

This works in tandem with https://github.com/inertiajs/inertia/pull/1617. It works by extending the Lazy feature but also informs the front-end on page load which deferred properties should be loaded automatically. These properties will then automatically trigger a reload with those properties automatically added in the `only` attribute.

This reduces the amount of front-end code required to pull this off. To add a long-running API query to be deferred, it would look like this:

```php
return Inertia::render('Organizations/Index', [
            'filters' => Request::all('search', 'trashed'),
            'organizations' => Inertia::defer(
                fn () =>
                Auth::user()->account->organizations()
                ->orderBy('name')
                ->filter(Request::only('search', 'trashed'))
                ->paginate(10)
                ->withQueryString()
                ->through(fn ($organization) => [
                    'id' => $organization->id,
                    'name' => $organization->name,
                    'phone' => $organization->phone,
                    'city' => $organization->city,
                    'deleted_at' => $organization->deleted_at,
                ])
            ),
        ]);
```

These deferred properties will only ever be loaded automatically after page loads/navigations. Otherwise, they behave just the same as Lazy props.


jbrooksuk (approved):

samehdoush:
It's great, waiting for its acceptance, so we can start using it immediately 👍  💯

adrum:
I just resolved the merge conflicts associated with this PR. Let me know if we want/need anything else.

adrum:
I'm having difficulty reproducing the failed test locally. 🤔 Anything I should try?

tests/ResponseTest.php
```diff
@@ -272,6 +273,24 @@ public function test_lazy_props_are_not_included_by_default(): void
...
+            return 'A defer value';
+        });
+
+        $response = new Response('Users', ['users' => [], 'defer' => $deferProp], 'app', '123');
+        $response = $response->toResponse($request);
+        $page = $response->getData();
+
+        $this->assertSame([], $page->props->users);
+        $this->assertSame(['defer'], $page->deferProps);
+        $this->assertObjectNotHasAttribute('defer', $page->props);
```
rojtjo, line 291:
```suggestion
        $this->assertFalse(property_exists($page->props, 'defer'));
```

tests/ResponseTest.php
```diff
@@ -291,6 +310,25 @@ public function test_lazy_props_are_included_in_partial_reload(): void
...
+
+        $deferProp = new DeferProp(function () {
+            return 'A deferred value';
+        });
+
+        $response = new Response('Users', ['users' => [], 'defer' => $deferProp], 'app', '123');
+        $response = $response->toResponse($request);
+        $page = $response->getData();
+
+        $this->assertObjectNotHasAttribute('users', $page->props);
```
rojtjo, line 328:
```suggestion
        $this->assertFalse(property_exists($page->props, 'users'));
```

rojtjo:
It looks like the `assertObjectNotHasAttribute` was removed in v10 and `assertObjectNotHasProperty` was added as a replacement in v10.1 and v9.6.

Easiest solution is to just use `assertFalse` in combination with a `property_exists` check.

See: https://github.com/sebastianbergmann/phpunit/issues/5478

adrum:
@rojtjo Thank you for getting to the bottom of that! I did not dig far enough into the error message to recognize the removal of those methods from PHPUnit.

chuck-waters:
This will solve so many things for us. Great addition!

reinink:
@adrum Really appreciate you sharing this idea, but I'm going to close this one for now for the reasons stated in your other PR:

https://github.com/inertiajs/inertia/pull/1617#issuecomment-2133398269

Once we've figured out an async story in Inertia I'd plan to loop back to this 👍

PR #534, opened 2023-08-23
Provide helper for asserting redirection

deleugpn:
Inertia uses 409 as the status code for redirection. When writing tests in Laravel with 

```
$this->post('/', headers: ['X-Inertia' => true])->assertRedirect('/redirected');
```

The test will fail because Laravel relies on `$this->isRedirect()` to validate whether the request returned a redirect. The `isRedirect()` method belongs to Symfony and it doesn't consider 409 as a valid status code, which means this always fail.

### DIFF
--- a/src/Testing/TestResponseMacros.php
+++ b/src/Testing/TestResponseMacros.php
@@ -26,6 +26,23 @@ class TestResponseMacros
         };
     }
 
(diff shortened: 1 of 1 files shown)


driesvints:

PR #537, opened 2023-08-27
Replace `array_merge` with spread operator in `HandleInertiaRequests`

osbre:


timacdonald:
This won't work on all version of PHP that Inertia currently supports.

Inertia currently supports 7.2+. The spread operator was introduced in PHP 7.4

osbre:
Any ideas why this package still supports Laravel 8 and PHP 7.3? They both reached EOL.

driesvints:
Not really. I think those can be removed soon.

PR #539, opened 2023-08-31
Restore request and session on redirects

timacdonald:
If there is an existing session on the redirect, it would be nice to restore it.

```php
Inertia::location(
    redirect()->to('/login')
)->with('foo', 'bar');
```

Also restoring the request while I'm in there.

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -109,14 +109,10 @@ class ResponseFactory
      */
     public function location($url): SymfonyResponse
     {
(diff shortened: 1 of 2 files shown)


rojtjo:
Wouldn't it make more sense to just return the original `RedirectResponse` instead of creating a new one? That way any other custom value such as the status code would also be preserved.

timacdonald:
@rojtjo yes! yes it would. Was too close to the project. Changed the approach.

PR #543, opened 2023-09-05
Support same PHP versions as Laravel 6.x

QuentiumYT:
With this requirement, Inertia Laravel will be satisfied by the same PHP version as Laravel >6. It also allows new developers to use PHP 8.3 (RC currently)

`inertiajs/inertia-laravel v0.6.9 requires php ^7.2|~8.0.0|~8.1.0|~8.2.0 -> your php version (8.3.0RC1) does not satisfy that requirement.`


QuentiumYT:
Well #540 might be more interesting by just adding the PHP 8.3 version

jessarcher:
Thanks! Given that PHP doesn't follow semver, I think it's probably safer to go with #540.

Issue #548, opened 2023-09-20
Add support for bun in `php artisan inertia:start-ssr`


Harrald:
maybe make it a config item to choose between node/bun.

reinink:
This has been added in #552 👍

PR #552, opened 2023-10-16
Add option for using the bun js runtime in SSR

smitssjors:
Addresses #548.

This adds a `--runtime` option which  can either be `node` or `bun`. The default is `node`.


reinink:
Hey sweet, nice addition, thanks! 🤙

PR #555, opened 2023-11-10
Update StartSsr.php to support rutime options

nikosv:
This commit allows to start the runtime with specific options. For example by running
`php artisan inertia:start-ssr --runtime_options=--max-old-space-size=4096  `
one can explicitly configure the memory limit and execute something like:
`node --max-old-space-size=4096 bootstrap/ssr/ssr.mjs`


driesvints:

PR #559, opened 2023-11-22
Add _ide_helpers.php to git export-ignore

gerardroche:
This file actually interferes with application-level LSP, because LSP
picks it up and indexes it, then things break like goto definition and
results in warnings about missing methods like `Route::get()`.


driesvints:
Done, thanks.

PR #566, opened 2023-12-02
Page URL: Fix handling of base url and reverse proxy forward prefix

flexponsive:
The `url` property of of the inertia HTTP response should contain the *user-facing request URI*, i.e. the absolute URL path and if applicable the query string.

Inertia currently has a bug where, if the laravel application has base url different from root, the `url` property contains the baseUrl twice (https://github.com/inertiajs/inertia-laravel/pull/372, https://github.com/inertiajs/inertia-laravel/issues/359, https://github.com/inertiajs/inertia-laravel/issues/421). This was introduced as a side-effect of a fix for a separate issue when the laravel application is deployed behind a reverse proxy, and where the reverse proxy has a forward prefix (https://github.com/inertiajs/inertia-laravel/pull/333).

This PR fixes the handling of the baseUrl case and preserves the behavior in case of reverse proxy with forward prefix. It also adds test cases for both cases.

## Example: Base URL duplicated by Inertia response 

There is a thread on laracasts, [URL duplicates in shared hosting deployed Inertia App](https://laracasts.com/discuss/channels/inertia/url-duplicates-in-shared-hosting-deployed-inertia-app), which illustrates the base url problem:

> However, the URL in the browser is rewritten whenever I access a page.
>
>https://www.mydomainname.com/project turns into https://www.mydomainname.com/project/project
>
>The same thing happens with the other route. https://www.mydomainname.com/project/github turns into https://www.mydomainname.com/project/project/github
>
> Checking with vue devtools, Inertia.initialPage.url seems to be set to the incorrect value (the one with a duplicate)

This problem happens because the `$request->getRequestUri()` contains the full request URI as sent to the application server, which also includes the base url:
https://github.com/inertiajs/inertia-laravel/blob/12e7937e99f6bdb114541c7f907bef1339b1c843/src/Response.php#L102

There are more users who experienced this problem, e.g. [root directory is displayed twice in URL on load and reload](https://laracasts.com/discuss/channels/laravel/root-directory-is-displayed-twice-in-url-on-load-and-reload?page=1&replyId=791970)

## Example: Reverse Proxy Deployment with Forward Prefix

Suppose the application has a public URL `https://intranet.example.org/to/project`. This is served through a reverse proxy, which requests the content in the background from `http://localhost:8080`. 

Using the `X-Forwarded-Prefix` header, the reverse proxy can [instruct symfony to prepend the forward prefix to the baseUrl](https://github.com/symfony/symfony/blob/b6ae3aa720ded594dea55d1db15fed9c03d97352/src/Symfony/Component/HttpFoundation/Request.php#L880). This was the setting reported in https://github.com/inertiajs/inertia-laravel/pull/333.

Since the forward prefix is not included in the `getRequestUri()` the problem of a duplicated base url does not arise in this case.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -99,7 +99,7 @@ class Response implements Responsable
         $page = [
             'component' => $this->component,
             'props' => $props,
-            'url' => $request->getBaseUrl() . $request->getPathInfo() . ($request->getQueryString() ? '?' . $request->getQueryString() : ''),
+            'url' => str_replace($request->getSchemeAndHttpHost(), '', $request->getUri()),
             'version' => $this->version,
         ];
 
--- a/src/Response.php
+++ b/src/Response.php
@@ -99,7 +99,7 @@ class Response implements Responsable
         $page = [
             'component' => $this->component,
             'props' => $props,
-            'url' => $request->getBaseUrl().$request->getRequestUri(),
+            'url' => $request->getBaseUrl() . $request->getPathInfo() . ($request->getQueryString() ? '?' . $request->getQueryString() : ''),
             'version' => $this->version,
         ];
 
--- a/tests/ResponseTest.php
+++ b/tests/ResponseTest.php
@@ -384,9 +384,15 @@ class ResponseTest extends TestCase
 
     public function test_page_url_with_proxy_forwarded_prefix() : void
     {
-        Request::setTrustedProxies([
-            '8.8.8.8'
-        ], Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_PREFIX);
+        try {
+            Request::setTrustedProxies([
+                '8.8.8.8'
+            ], Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_PREFIX);
+        } catch (\Throwable $th) {
+            $this->markTestSkipped('Trusted proxies not supported in this version of Symfony');
+            return;
+        }
+
         $request = Request::create('/user/123', 'GET', [], [], [], [
             'SCRIPT_FILENAME' => '/ws/test/app-base-url/public/index.php',
             'SCRIPT_NAME' => '/index.php',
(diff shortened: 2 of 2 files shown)


flexponsive:
## Related PRs

I saw there are a couple of open pull requests here addressing the same issue:

- https://github.com/inertiajs/inertia-laravel/pull/446 and https://github.com/inertiajs/inertia-laravel/pull/422 both propose reverting https://github.com/inertiajs/inertia-laravel/pull/333 and setting `'url' => $request->getRequestUri()`. This would cause the issue with the forward prefix in the reverse proxy setting to re-appear.

- https://github.com/inertiajs/inertia-laravel/pull/437 suggests `'url' => $request->fullUrl()` which solves both the base url and forward prefix issues, but as a side effect changes the `url` property to include the full URL, including protocol and host part. This causes other tests to fail.

- Finally, https://github.com/inertiajs/inertia-laravel/pull/360 suggests basically the same change that I am proposing here (`'url' => $request->getBaseUrl().$request->getPathInfo().$queryString`), which works both the base url and forward prefix cases. It is only missing a test case for the forward prefix case.

## User impact

Additionally to the examples above, the "inertia duplicates the base url" problem also came up in different forms on stackoverflow, e.g. [Ziggy router modifying browser urls](https://stackoverflow.com/questions/75666272/ziggy-router-modifying-browser-urls) or [root directory is displayed twice in URL on load and reload](https://stackoverflow.com/questions/71943581/root-directory-is-displayed-twice-in-url-on-load-and-reload)

Many thanks to the entire Inertia team for this amazing project! I hope this PR can contribute to solving one of the very few longstanding issues.

src/Response.php
```diff
@@ -99,7 +99,7 @@ public function toResponse($request)
         $page = [
             'component' => $this->component,
             'props' => $props,
-            'url' => $request->getBaseUrl().$request->getRequestUri(),
```
RobertBoes, line 102: I think this can be changed to just using `fullUrl()`, https://github.com/laravel/framework/blob/10.x/src/Illuminate/Http/Request.php#L116-L123

```suggestion
            'url' => $request->fullUrl(),
```

RobertBoes (changes requested):

src/Response.php
flexponsive: `fullUrl` correctly takes care of the base url and forward prefix, but it also prepends the protocol and hostname. This may have an impact on the inertia clients for the different JavaScript frameworks. It causes 10 failures in the `ResponseTest`, for example:

```
There were 10 failures:

1) Inertia\Tests\ResponseTest::test_server_response
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'/user/123'
+'http://localhost/user/123'

/ws/dev/inertia-laravel/tests/ResponseTest.php:46
/ws/dev/inertia-laravel/vendor/orchestra/testbench-core/src/PHPUnit/TestCase.php:42

....
```

The advantage of using fullUrl is that it would be more consistent, since the prop is already called `url` but in reality only contains the path URI. But it is a potentially breaking change if JS clients assume they receive only the path URI.

@RobertBoes what do you prefer?

murilomagalhaes:
At this point i've just given up for any PR that fixes this problem to be merged. Sadly, i've seen a lot of them being closed or just ignored.

src/Response.php
RobertBoes: Hmm yeah, you're right, didn't think about that

> Also, detecting the protocol (https?) and actual client facing hostname can be a headache in a reverse proxy environment...

It shouldn't be, if the correct headers are sent and the `TrustProxies` middleware is set up properly (which is what you should do anyways if you're using a reverse proxy)

flexponsive:
Following the discussion with @RobertBoes, I changed the code to make explicit that - in line with the [inertia protocol example](https://inertiajs.com/the-protocol#inertia-responses) - the `url` property should not contain the HTTP scheme and host name.

The code now reads: 

```php
 'url' => str_replace($request->getSchemeAndHttpHost(), '', $request->getUri()),
```

I would have liked use the `fullUrl` method, but this runs into the problem that when accessing the root URL, the `fullUrl` method does not return a trailing slash (so one would have to write `str_replace($request->getSchemeAndHttpHost(), '', $request->fullUrl()) ?: '/'`, which is a bit ugly).

jessarcher:
Thanks for the PR, and I'm sorry for the delay! I ended up going with a different but very similar approach (#592), but this PR was very helpful!

taomuza201:
laravel docker nginx subpath

i fix  Response.php   `  'url' => Str::after($request->fullUrl(), $request->getSchemeAndHttpHost()),`

on index page is work buy  click link login url is lost
image


image

PR #571, opened 2023-12-15
Get SSR endpoint from ENV

wispoz:
This pull request add env variable for Inertia SSR, this will ne useful change endpoist without publish inertia config.


joetannenbaum:
This is now implemented  in 2.x, closing.

PR #582, opened 2024-01-23
feat: validation errors can return all error messages

squiaios:
Right now, when you have errors with multiple messages, only the first is returned as string.

I added a function `returnOnlyFirstErrorMessage` in middleware in order to be able to overide on our app if we want to display all the messages in an array


driesvints:
Tests fail here

PR #595, opened 2024-03-08
Testing lazy props

JeroenGovers:
I added the method `requestProp` (name open for discussion 😉) to the AssertableInertia class which let you easy test LazyProps.

```php
/** @test */
public function test_props(): void
{
    $this
        ->get('/')
        ->assertInertia(function (AssertableInertia $page) {
            $page->has('normalPropKey');
            $page->missing('lazyPropKey');

            $page->requestProp('lazyPropKey', function (AssertableInertia $page) {
                $page->where('lazyPropKey', '1');
            });
        });
}
```

The method uses the previous request, applies the ```X-Inertia-Partial-Component``` and ```X-Inertia-Partial-Data``` headers, which are automatically seeded and resends it. A convenient way for the tester, and they do not need to know the internal mechanics of Intertia. Which I found in [#604](https://github.com/inertiajs/inertia/discussions/604).

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -2,6 +2,8 @@
 
 namespace Inertia;
 
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Foundation\Http\Events\RequestHandled;
 use LogicException;
 use Inertia\Ssr\Gateway;
 use ReflectionException;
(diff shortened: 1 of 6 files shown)


pascalbaljet:
Thanks for this! I've reworked it a bit in #738 as this was targeted at 0.x. I also didn't want to rely on the app environment and binding in the container. When the new PR is merged, I'll make sure to add you as a contributor!

JeroenGovers:
> Thanks for this! I've reworked it a bit in #738 as this was targeted at 0.x. I also didn't want to rely on the app environment and binding in the container. When the new PR is merged, I'll make sure to add you as a contributor! 

@pascalbaljet it's great to hear that the new test helpers are now available in 2.x. I will definitely use them in my tests. You spoke of making me a contributor when the PR is merged. I would realy love that ☺️

Issue #93, opened 2020-01-03
[0.2] [bug?] A single resource has its properties nested in a data attribute.


johanvanhelden:
Since the upgrade to 0.2 a resource collection has a data property that contains the collection of resources. I get that. That is awesome.

What I don't understand, and I hope this is a bug, is why this also applies to a single resource.

For example:
```php
return Inertia::render('Home', [
	'projects' => ProjectResource::collection($projects),	
```
Gives me a Vue object like this:
![Screenshot from 2020-01-03 15-34-10](https://user-images.githubusercontent.com/19389981/71728845-8906dd80-2e3e-11ea-9471-2e8e5af390a8.png)

Again, makes complete sense to me! You just create a loop over `project.data` in Vue and can simply use `{{ project.name }}` to render data for a single project.

But, this example:
```php
return Inertia::render('Project/Show', [
    'project' => new ProjectResource($project),
```
Gives me a Vue object like this:
![Screenshot from 2020-01-03 15-35-36](https://user-images.githubusercontent.com/19389981/71728899-b6538b80-2e3e-11ea-8989-c67c6f477dff.png)

And here I would have expected the old behavior. A project object with the properties directly inside  it and not nested within a data property. Because now I would have to rewrite everything from
`{{ project.name }}` to `{{ project.data.name }}`.

So for a single project inside a project collection I can use `{{ project.name }}` but for the single resource I need to use `{{ project.data.name }}`. That is just not consistent.

So again, I would just like to make sure that this is the intended behavior for a single resource. It just looks to much like a bug to me.

reinink:
This is simply how API resources in Laravel work. Even on single resources, it nests them within the `data` property. You can disable this behaviour using `Resource::withoutWrapping()`. See [here](https://laravel.com/docs/6.x/eloquent-resources#data-wrapping) for more info.

PR #619, opened 2024-05-10
[2.x] Automatically disable wrapping for API resource props

TheDoctor0:
Continuation of #432 as it remains a problem in the latest version of Inertia.

I was in the process of porting the project to Inertia and I encountered a very unexpected issue with the JSON resources serialization. Some had a different structure in comparison to the one returned from API routes - a data wrapper was present.

I tinkered with the source code and discovered that the issue lies in how resources are being serialized in the response [master/src/Response.php#L136](https://github.com/inertiajs/inertia-laravel/blob/master/src/Response.php?rgh-link-date=2022-07-19T10%3A51%3A48Z#L136).

I was surprised to find out that the FakeResource used for tests has a $wrap property set to null to mitigate this issue but that has not been documented anywhere. Without this property, the test for response with JSON resource fails.

To fix this unexpected behavior I changed the logic for resource serialization to make sure not to use the data wrapper.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -226,7 +226,9 @@ class Response implements Responsable
             }
 
             if ($value instanceof ResourceResponse || $value instanceof JsonResource) {
-                $value = $value->toResponse($request)->getData(true);
+                $value = tap($value, static function ($value) {
+                    $value->withoutWrapping();
+                })->toResponse($request)->getData(true);
             }
 
             if (is_array($value)) {
--- a/tests/ResponseTest.php
+++ b/tests/ResponseTest.php
@@ -14,6 +14,7 @@ use Illuminate\View\View;
 use Inertia\AlwaysProp;
 use Inertia\LazyProp;
 use Inertia\Response;
+use Inertia\Tests\Stubs\FakeNestedResource;
 use Inertia\Tests\Stubs\FakeResource;
 use Mockery;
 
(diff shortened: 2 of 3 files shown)


reinink:
Hey thanks for this contribution!

This issue has come up in the past, see #93.

My concern with this change is that it's a pretty big breaking change that would require updating your page components to use the non-wrapped version of the data.

To this point I've recommended that people using resources put `JsonResource::withoutWrapping` in their app service provider, and that way folks can opt-in to this behavior by using this existing Laravel option.

We could bake this opinion into this adapter directly, but (at least from what I can tell) there would be no way to opt-out of it, so we need to make sure that's the right decision.

Unfortunately I don't use resources in my Inertia apps, so I can't really speak to this. I'm going to ask around to see what others think 👍

Either way it would have to be included as part of a major version release, which we are planning to do soon, so if we do want to make this change the timing is right.

LucaRed:
I use JSON resources in my Inertia projects all the time. Why? So I can filter out data that I don't want to share publicly *and* add more properties at the same time (e.g. relations, computed attributes, etc...), all from a centralized location that I can reuse.

Personally, I've always found the `data` wrapping annoying, but I've never disabled it because of another annoyance: the `$wrap` property on JsonResources can be set to `null` to prevent wrapping, but it doesn't work when you instantiate a collection like this: `ExampleResource::collection($models)`.

Related issues/PRs: 
- https://github.com/laravel/framework/issues/49155
- https://github.com/laravel/framework/pull/43463

So, I'm all for disabling the JSON `data` wrapping for the next major release 👍 

I understand that some people may not want to change all the references in their projects,  so could we add a global configuration to revert back to the old `data`-wrapping behavior?

reinink:
@TheDoctor0 Hey I think we're going to make this change for 2.x. Would you mind resolving the conflicts and also adding a test for this (assuming it's not too hard to test)?

TheDoctor0:
@reinink of course, I will update the PR whenever I have some free time this week.

reinink:
@TheDoctor0 great, thanks!

TheDoctor0:
@reinink change ported to 2.x version and covered with tests.

pascalbaljet:
You can achieve this behavior by calling the static `JsonResource::withoutWrapping()` method, for example, in a Service Provider ([docs](https://laravel.com/docs/12.x/eloquent-resources#data-wrapping)). This works for both (non-paginated) collections and individual resources.

To make sure it works as your proposed change, I reverted your changes in `Inertia\Response`, and called the `withoutWrapping` method before your new tests, and they all passed.

PR #622, opened 2024-05-14
Exclude `except` props from partial reloads

lepikhinb:
The PR introduces an ability to exclude properties from partial responses. The `except` option can be used together with `only` to refine nested props.

The corresponding client-side PR: https://github.com/inertiajs/inertia/pull/1876

```js
router.visit(url, {
  only: ['auth'],
  except: ['auth.user'],
})
```

```php
$props = [
    // Included on partial reload
    'auth' => [
        // Excluded from partial reload
        'user' => new LazyProp(function () {
            return [
                'name' => 'Jonathan Reinink',
                'email' => 'jonathan@example.com',
            ];
        }),
        // Included on partial reload
        'refresh_token' => 'value',
    ],
    // Excluded from partial reload
    'shared' => [
        'flash' => 'value',
    ],
];
```

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -111,7 +111,7 @@ class Response implements Responsable
     {
         $isPartial = $request->header(Header::PARTIAL_COMPONENT) === $this->component;
 
-        if(!$isPartial) {
+        if(! $isPartial) {
             $props = array_filter($this->props, static function ($prop) {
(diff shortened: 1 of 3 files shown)


driesvints:
Please make sure to mark this as ready for review if you need one.

lepikhinb:
@driesvints just want to get the client PR ready first

PR #623, opened 2024-05-15
Add Defer Prop

adrum:
Recreating this [PR](https://github.com/inertiajs/inertia-laravel/pull/531) due to the master branch deletion. If the previous PR was closed on purpose, I apologize for the noise. 

Original description: I felt inspired by Livewire 3 Lazy feature which automatically fetched the data on page load. I think there's a ton of value in Inertia's current Lazy prop feature, but I think there are also scenarios where adding a Defer prop feature which automatically loads Lazy data on page load.

This works in tandem with https://github.com/inertiajs/inertia/pull/1617. It works by extending the Lazy feature but also informs the front-end on page load which deferred properties should be loaded automatically. These properties will then automatically trigger a reload with those properties automatically added in the `only` attribute.

This reduces the amount of front-end code required to pull this off. To add a long-running API query to be deferred, it would look like this:

```php
return Inertia::render('Organizations/Index', [
            'filters' => Request::all('search', 'trashed'),
            'organizations' => Inertia::defer(
                fn () =>
                Auth::user()->account->organizations()
                ->orderBy('name')
                ->filter(Request::only('search', 'trashed'))
                ->paginate(10)
                ->withQueryString()
                ->through(fn ($organization) => [
                    'id' => $organization->id,
                    'name' => $organization->name,
                    'phone' => $organization->phone,
                    'city' => $organization->city,
                    'deleted_at' => $organization->deleted_at,
                ])
            ),
        ]);
```

These deferred properties will only ever be loaded automatically after page loads/navigations. Otherwise, they behave just the same as Lazy props.


reinink:
@adrum Really appreciate you sharing this idea, but I'm going to close this one for now for the reasons stated in your other PR:

https://github.com/inertiajs/inertia/pull/1617#issuecomment-2133398269

Once we've figured out an async story in Inertia I'd plan to loop back to this 👍

PR #496, opened 2023-01-20
Add `$request->inertia()` IDE helper

ycs77:
Added the `$request->inertia()` IDE helper

PR #625, opened 2024-05-15
[1.x] Add `$request->inertia()` IDE helper

ycs77:
**Re-open #496**

---

Add `$request->inertia()` IDE helper.

And this PR re-export the `_ide_helpers.php` file, if no export the developer can't look the IDE helpers when developing.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,7 +7,6 @@
 /.gitignore export-ignore
 /.github export-ignore
 /.php-cs-fixer.dist.php export-ignore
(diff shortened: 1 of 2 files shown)


BeingTomGreen:
This update has caused an issue in both PHPStorm and VSCode, because there are now multiple definitions the of the Illuminate\Http\Request class.

I don't use IDE helper.

Weirdly I don't seem to get this for TestResponse, although I'm not really sure why.

Issue #176, opened 2020-10-29
Custom `rootView` set in `HandleInertiaRequests` not taken into account for errors


rgehan:
The new `HandleInertiaRequests` middleware exposes a `rootView` property, allowing to set a custom root view.

It works fine for regular `Inertia::render` in controllers, but doesn't seem to be taken into account when [returning custom error views, as the doc suggests](https://inertiajs.com/error-handling#production).

In my case, this happens for a 404 error. I imagine Laravel handles 404 errors differently.
The request might not go through the middleware stack.

I just realized that it wouldn't send the shared data to the error page either.

Here's a stack trace:
```
[2020-10-29 15:26:42] local.ERROR: View [app] not found. {"exception":"[object] (InvalidArgumentException(code: 0): View [app] not found. at /<project>/vendor/laravel/framework/src/Illuminate/View/FileViewFinder.php:137)
[stacktrace]
#0 /<project>/vendor/laravel/framework/src/Illuminate/View/FileViewFinder.php(79): Illuminate\\View\\FileViewFinder->findInPaths('app', Array)
#1 /<project>/vendor/laravel/framework/src/Illuminate/View/Factory.php(138): Illuminate\\View\\FileViewFinder->find('app')
#2 /<project>/vendor/laravel/framework/src/Illuminate/Routing/ResponseFactory.php(85): Illuminate\\View\\Factory->make('app', Array)
#3 /<project>/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php(261): Illuminate\\Routing\\ResponseFactory->view('app', Array)
#4 /<project>/vendor/inertiajs/inertia-laravel/src/Response.php(93): Illuminate\\Support\\Facades\\Facade::__callStatic('view', Array)
#5 /<project>/app/Exceptions/Handler.php(51): Inertia\\Response->toResponse(Object(Illuminate\\Http\\Request))
#6 /<project>/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(416): App\\Exceptions\\Handler->render(Object(Illuminate\\Http\\Request), Object(InvalidArgumentException))
#7 /<project>/vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(114): Illuminate\\Foundation\\Http\\Kernel->renderException(Object(Illuminate\\Http\\Request), Object(InvalidArgumentException))
[...]

PR #626, opened 2024-05-22
Change the rootView from configuration

hafijul233:
`ResponseFactory.php` class defines the *$rootView* point to `app.blade.php`. I thought it would be easy to configure it. Anyone who publishes the config file can easily change it. `ServiceProvider` boot method workaround also has the same output. IMHO This way, it feels more neat.

image

### DIFF
--- a/config/inertia.php
+++ b/config/inertia.php
@@ -4,7 +4,7 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Server Side Rendering
+    | Default Root View
     |--------------------------------------------------------------------------
     |
     | This options configures which view should be set as default root view.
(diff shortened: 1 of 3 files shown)


NickSdot:
Inertia intentionally did choose to be configured via middleware, instead of config. There are considerations ([for other reasons](https://github.com/inertiajs/inertia-laravel/issues/176#issuecomment-719485424)) to go with a ServiceProvider instead.

However, you already can use `Inertia::setRootView('custom.layout');` wherever you want. Like, globally in AppServiceProvider or for instance in a specific controller.

Personally, not sure if we really need yet another way to configure this.

PR #627, opened 2024-05-25
Add "always" props using new `Inertia::always()` wrapper

lepikhinb:
Introduce `Inertia::always()` wrapper to replace the current [persistent props](https://github.com/inertiajs/inertia-laravel/pull/621) implementation.

```php
return [
    'flash' => Inertia::always(session('flash')),
];
```

Lazy data evaluation behavior:

```php
    // ALWAYS included on first visit...
    // OPTIONALLY included on partial reloads...
    // ALWAYS evaluated...
    'users' => User::get(),

    // ALWAYS included on first visit...
    // OPTIONALLY included on partial reloads...
    // ONLY evaluated when needed...
    'users' => fn () => User::get(),

    // NEVER included on first visit...
    // OPTIONALLY included on partial reloads...
    // ONLY evaluated when needed...
    'users' => Inertia::lazy(fn () => User::get()),
    
    // ALWAYS included on first visit...
    // ALWAYS included on partial reloads...
    // ALWAYS evaluated...
    'users' => Inertia::always(User::get()),
]);
```

### DIFF
--- a/src/Inertia.php
+++ b/src/Inertia.php
@@ -12,7 +12,7 @@ use Illuminate\Support\Facades\Facade;
  * @method static void version(\Closure|string|null $version)
  * @method static string getVersion()
  * @method static \Inertia\LazyProp lazy(callable $callback)
- * @method static \Inertia\AlwaysProp always(callable $callback)
+ * @method static \Inertia\AlwaysProp always(mixed $value)
  * @method static \Inertia\Response render(string $component, array|\Illuminate\Contracts\Support\Arrayable $props = [])
  * @method static \Symfony\Component\HttpFoundation\Response location(string|\Symfony\Component\HttpFoundation\RedirectResponse $url)
  * @method static void macro(string $name, object|callable $macro)
--- a/src/Inertia.php
+++ b/src/Inertia.php
@@ -12,15 +12,13 @@ use Illuminate\Support\Facades\Facade;
  * @method static void version(\Closure|string|null $version)
  * @method static string getVersion()
  * @method static \Inertia\LazyProp lazy(callable $callback)
+ * @method static \Inertia\AlwaysProp always(callable $callback)
  * @method static \Inertia\Response render(string $component, array|\Illuminate\Contracts\Support\Arrayable $props = [])
  * @method static \Symfony\Component\HttpFoundation\Response location(string|\Symfony\Component\HttpFoundation\RedirectResponse $url)
  * @method static void macro(string $name, object|callable $macro)
  * @method static void mixin(object $mixin, bool $replace = true)
  * @method static bool hasMacro(string $name)
  * @method static void flushMacros()
- * @method static void persist(string|array|\Illuminate\Contracts\Support\Arrayable $props)
- * @method static array getPersisted()
- * @method static void flushPersisted()
  *
  * @see \Inertia\ResponseFactory
  */
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -53,9 +53,7 @@ class Middleware
     public function share(Request $request)
     {
         return [
-            'errors' => new AlwaysProp(function () use ($request) {
-                return $this->resolveValidationErrors($request);
-            }),
+            'errors' => Inertia::always($this->resolveValidationErrors($request)),
         ];
     }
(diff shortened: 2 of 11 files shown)


src/Middleware.php
```diff
@@ -60,9 +53,9 @@ public function version(Request $request)
     public function share(Request $request)
     {
         return [
-            'errors' => function () use ($request) {
+            'errors' => new AlwaysProp(function () use ($request) {
```
reinink: Wondering if we should use `Inertia::always()` here instead of `AlwaysProp`? Just because I could see people viewing the source of this file often and I think we want to encourage `Inertia::always()` over `AlwaysProp`.

src/ResponseFactory.php
```diff
@@ -115,6 +88,11 @@ public function lazy(callable $callback): LazyProp
         return new LazyProp($callback);
     }
 
+    public function always(callable $callback): AlwaysProp
```
reinink: Given that "always" props are always evaluated, there isn't really any reason to use a closure. So I think we should accept anything here like a regular prop:

```php
'authName' => Inertia::always(Auth::user()->name),
```

That said, it would be nice to still allow using a closure, just because that's sometimes nice for code formatting, so I think we should still support that as well.

src/Middleware.php
reinink: Combined with the other change I'm suggesting, this could actually just be:

```php
'errors' => Inertia::always($this->resolveValidationErrors($request)),
```

lepikhinb:
@reinink made `AlwaysProp` accept any values including callable 👌🏻

src/Inertia.php
```diff
@@ -12,15 +12,13 @@
  * @method static void version(\Closure|string|null $version)
  * @method static string getVersion()
  * @method static \Inertia\LazyProp lazy(callable $callback)
+ * @method static \Inertia\AlwaysProp always(callable $callback)
```
reinink: Needs to be updated for new supported types.

```suggestion
 * @method static \Inertia\AlwaysProp always(mixed $value)
```

reinink:
@lepikhinb Hey thanks for putting this together! 💪

gabrielrbarbosa:
@lepikhinb "errors" prop is empty with laravel validation after this update

src/Middleware.php
```diff
@@ -60,9 +53,7 @@ public function version(Request $request)
     public function share(Request $request)
     {
         return [
-            'errors' => function () use ($request) {
-                return $this->resolveValidationErrors($request);
-            },
+            'errors' => Inertia::always($this->resolveValidationErrors($request)),
```
gabrielrbarbosa, line 56: this is a breaking change, isnt it? @lepikhinb

datlechin:
After updating to v1.3.0, no errors are included in the response; it is always empty.

This issue did not occur in v1.2.0.

@reinink, could you please look into this?

lepikhinb:
@datlechin @gabrielrbarbosa I couldn't reproduce the issue. Is there a chance you can make a quick reproduction?

src/Middleware.php
lepikhinb: Shouldn't be

datlechin:
@lepikhinb Create a route that performs data validation. Send a request to this route with invalid data and verify that the response does not contain any error messages.

lepikhinb:
@datlechin Please make a repo with a complete reproduction.

Here's PingCRM with Inertia 1.3.

CleanShot 2024-06-22 at 18 36 43@2x

datlechin:
Ah i see that what happened

Previously, I used the `appendTo` method to add Inertia’s middleware, which was works with versions prior to v1.3.0:

```php
->withMiddleware(function (Middleware $middleware) {
    $middleware->appendTo(HandleInertiaRequests::class);
})
```

After updating to v1.3.0, the approach needs to be changed. Now, you must use the `web` method to add Inertia’s middleware.

gabrielrbarbosa:
@lepikhinb I found out what happened here. We had previously overridden the Response class which made this new update incompatible, I will handle this here or just fix at 1.2.0 for now, thanks anyway!

Issue #633, opened 2024-05-30
adding to HandleInertiaRequests->share array no longer works in version 1.2.0


Mrkars:
I have this code block that use to work in version 1.0.0. It is no longer working in version 1.2.0. It use to return the auth.user along with permissions. Now it appears that permissions is just overwriting the entire array and that is all that is returned.

//HandleInertiaRequests.php
```
public function share(Request $request): array
{
      return array_merge(parent::share($request), [
            'auth.user.permissions' => fn () => $request->user()->getAllPermissions(),
      ]);
}
```

driesvints:
@lepikhinb could this be related to the changes you made?

lepikhinb:
@Mrkars can you share how does the original entire array look?

Mrkars:
Sure thing. So in version 1.0.0, the array returned to the frontend looks like this:

{ 
"id": 1, "name": "User Name", "email": "email@email.com", "email_verified_at": "2024-05-15T19:29:47.000000Z", "two_factor_confirmed_at": null, "created_at": "2024-05-15T19:29:47.000000Z", "updated_at": "2024-05-15T19:35:48.000000Z", "deleted_at": null,
"permissions": 
[ { 
"id": 1, "name": "Configuration-Read", "guard_name": "web", "created_at": "2024-05-15T19:29:47.000000Z", "updated_at": "2024-05-15T19:29:47.000000Z", 
"pivot": 
{ "role_id": 1, "permission_id": 1 } 
}, 
{ 
"id": 2, "name": "Configuration-Edit", "guard_name": "web", "created_at": "2024-05-15T19:29:47.000000Z", "updated_at": "2024-05-15T19:29:47.000000Z", 
"pivot": { "role_id": 1, "permission_id": 2 } 
}, ], 
"roles": [ { "id": 1, "name": "Super Admin", "guard_name": "web", "created_at": "2024-05-15T19:29:47.000000Z", "updated_at": "2024-05-15T19:29:47.000000Z", "pivot": { "model_type": "App\\Models\\User", "model_id": 1, "role_id": 1 } } ], 
"two_factor_enabled": false 
}

In version 2.0.0 it then looks like this:

{ 
"permissions": 
[ { 
"id": 1, "name": "Configuration-Read", "guard_name": "web", "created_at": "2024-05-15T19:29:47.000000Z", "updated_at": "2024-05-15T19:29:47.000000Z", 
[...]

PR #631, opened 2024-05-29
[1.x] Fix resolving response arrayable properties

VicGUTT:
Prior to https://github.com/inertiajs/inertia-laravel/pull/620, property instances were resolved [before](https://github.com/inertiajs/inertia-laravel/pull/620/files?diff=split&w=0#diff-e1ee2fcc629f38d16eaba23b95a87eb8e7466fae0954483c4b8ca8ad78242e0aL136) handling arrayable/array values when sending a response.

Consider the following Inertia response dump:

<details>

<summary>AsArrayable</summary>

```php
use Illuminate\Contracts\Support\Arrayable;

class AsArrayable implements Arrayable {
    public function __construct(private array $data) {}

    public static function make(array $data): static {
        return new static($data);
    }

    public function toArray(): array {
        return $this->data;
    }
}
```
</details>

```php
use Inertia\Inertia;
use Illuminate\Http\Request;

$request = Request::create('/auth', 'GET');

$response = Inertia::render('auth', [
    'user' => fn () => [
        'full_name' => 'Victor Gutt',
        'organizations' => AsArrayable::make([
            fn () => [
                'name' => 'Transl.me',
            ],
        ]),
    ],
]);

$response = $response->toResponse($request);

dd(
    $response->original->getData()['page']['props'],
);
```

Before this fix, the dump would be similar to:

```php
Inertia\Tests\ResponseTest
array:1 [
  "user" => array:2 [
    "full_name" => "Victor Gutt"
    "organizations" => Inertia\Tests\Stubs\AsArrayable {#1291
      #data: array:1 [
        0 => Closure() {#1277
          class: "Inertia\Tests\ResponseTest"
          this: Inertia\Tests\ResponseTest {#173 …}
        }
      ]
    }
  ]
]
```

As we can see, nested _(`user.organizations`)_ arrayable values are no longer resolved.

After this fix, the dump is similar to:

```php
Inertia\Tests\ResponseTest
array:1 [
  "user" => array:2 [
    "full_name" => "Victor Gutt"
    "organizations" => array:1 [
      0 => array:1 [
        "name" => "Transl.me"
      ]
    ]
  ]
]
```

This PR aims to bring back the behaviour prior to #620.

PR #634, opened 2024-05-31
[1.x] Fix shared dot props

RobertBoes:
I _think_ #620 introduced a breaking change when sharing props through the middleware, as described in #633

A use-case would be the following:

```php
{
	{
		return array_merge(parent::share($request), [
			'user.verified' => true,
		]);
	}
}
```

Then render the page using:
```php
Inertia::render('User/Show', [
	'user' => [
		'name' => 'John',
	],
]);
```

This would result in just `props: { user: { name: 'John' }}` being sent to the frontend instead of `props: { user: { name: 'John', verified: true }}`, with this PR the latter would happen.

I attempted to fix the issue by merging the props instead of overwriting them. However, I'm not actually sure why this feature worked before, since it seems like props were always overwritten.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -137,7 +137,7 @@ class Response implements Responsable
     /**
      * Resolve all arrayables properties into an array.
      */
-    public function resolveArrayableProperties(array $props, Request $request, bool $unpackDotProps = true): array
+    public function resolveArrayableProperties(array $props, Request $request): array
     {
         foreach ($props as $key => $value) {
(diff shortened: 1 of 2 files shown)


VicGUTT:
Hey @RobertBoes, without looking to deep into it and although that was not the intended fix of my PR, I was wondering if my PR #631 doesn't also resolve the issue described here?

The point of my PR was to simply revert back the order of which property instances and arrayable properties were resolved prior to #620.

PR #636, opened 2024-06-01
[1.x] Fix resolving shared dot props

lepikhinb:
Based on https://github.com/inertiajs/inertia-laravel/pull/634

> I think https://github.com/inertiajs/inertia-laravel/pull/620 introduced a breaking change when sharing props through the middleware, as described in https://github.com/inertiajs/inertia-laravel/issues/633

This PR includes extra tests to ensure shared props can be included and excluded from partial responses.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -139,6 +139,8 @@ class Response implements Responsable
      */
     public function resolveArrayableProperties(array $props, Request $request, bool $unpackDotProps = true): array
     {
+        $dotProps = [];
+
         foreach ($props as $key => $value) {
             if ($value instanceof Arrayable) {
                 $value = $value->toArray();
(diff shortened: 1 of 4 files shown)


VicGUTT:
Same comment as in https://github.com/inertiajs/inertia-laravel/pull/634#issuecomment-2143360780.

Might be worth investigation if the PR https://github.com/inertiajs/inertia-laravel/pull/631 also solve this issue?

lepikhinb:
@VicGUTT it doesn't. However, I think #631 can be simplified a bit.

VicGUTT:
Alright, too bad. Thanks for confirming.

Sure, in what way? That PR simply reverted back the method "resolvePropertyInstances" to what it previously was

lepikhinb:
@reinink I rewrote the entire partial request resolving logic. It seems that the only way to support nested partial props without introducing any breaking changes is to check each prop independently to determine whether it should be included or excluded from the partial response.

Additionally, we need to resolve dot-notated props last to ensure they are merged **after** any parent closures are resolved.

PR #650, opened 2024-08-02
[1.x] Put Vite check before Mix

nshiro:
Now that most people use `Vite`, why don't we check `Vite` manifest file before `Mix`'s for performance?

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -32,11 +32,11 @@ class Middleware
             return md5(config('app.asset_url'));
         }
 
-        if (file_exists($manifest = public_path('mix-manifest.json'))) {
+        if (file_exists($manifest = public_path('build/manifest.json'))) {
             return md5_file($manifest);
         }
 
-        if (file_exists($manifest = public_path('build/manifest.json'))) {
+        if (file_exists($manifest = public_path('mix-manifest.json'))) {
             return md5_file($manifest);
         }
 


RobertBoes:
Don't think this would have any sort of significant impact, not to mention PHP caches a `file_exists` call, so for repeat calls this would even have less of an impact

nshiro:
I should have checked the performance first.
On my computer (Ryzen 7 3700X, PHP 8.3) when I ran the code below, the average of ten runs was 1.021 seconds.
But when I put a mix file check first (which doesn't exist), I got an average time of 0.892 seconds.
The difference is 0.129 seconds for 100,000 hits.
This may not be an insignificant impact. But it could be worth considering in certain contexts.

```php
<?php

$start = microtime(true);

for ($i=1; $i < 100_000; $i++) {
    if (file_exists($manifest = 'public/build/manifest.json')) {
        md5_file($manifest);
        continue;
    }

    if (file_exists($manifest = 'public/build/mix-manifest.json')) {
        md5_file($manifest);
        continue;
    }
}

$end = microtime(true);
$executionTime = $end - $start;
echo "Execution time: " . $executionTime . " seconds";
```

RobertBoes:
Is that with or without opcache? I'd imagine that call would benefit greatly from opcache.

nshiro:
Opcache is on.

Interestingly, the time didn't change when I checked only Vite and when I checked only Mix.
I cut out `md5_file($manifest)` and confirmed.
According to this page,
https://www.php.net/manual/en/function.clearstatcache.php
```
This function caches information about specific filenames, so you only need to call clearstatcache() if you are performing multiple operations on the same filename and require the information about that particular file to not be cached.
```
Because there was information that `specific filenames` referred to the last accessed file, I asked the AI (Claude) the following:
`Does PHP's file_exists function only cache the most recent call?`

The AI then answered as follows:

```
The file_exists() function in PHP does not cache results itself. However, PHP's internal file status cache, which is used by several filesystem functions including file_exists(), does cache information about files.
By default, this cache only stores information about the most recently accessed file. This means that repeated calls to file_exists() for the same file within a short time frame might use the cached result, but calls for different files will typically cause a new filesystem check.
```

If this answer is correct, it seems that the cache is not being used.

pascalbaljet:
I think this is a good change, but could you please change it to the 2.x branch or resubmit it? Thanks!

nshiro:
Thank you for the review.
I didn't consider this when creating the PR, but I now realize this could be a breaking change due to the behavior modification.
So I'll just close this PR and won't create a new one.
Thank you.

PR #652, opened 2024-08-09
Replace md5 with xxhash

RobertBoes:
I recently noticed laravel/framework#52301, where most of the md5 calls will be replaced with xxhash, which is much faster than md5. For Inertia's asset versioning we don't need it to be cryptographically correct, but a performance improvement would be nice, especially since the middleware is executed on every request.

I think it's save to change this in the Inertia adapter, since all it would do is force a reload, which would happen when updating assets anyway.

Support for xxhash was added to PHP 8.1 (https://php.watch/versions/8.1/xxHash), so we can safely use this algorithm instead of md5.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -29,15 +29,15 @@ class Middleware
     public function version(Request $request)
     {
         if (config('app.asset_url')) {
-            return md5(config('app.asset_url'));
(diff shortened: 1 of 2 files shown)


RobertBoes:
Going to target 2.x

PR #654, opened 2024-08-20
[1.x] remove IDE helpers from package

krzysztofrewak:
Since [v1.1.0](https://github.com/inertiajs/inertia-laravel/releases/tag/v1.1.0) release (with changes from #625) my IDE has a big problem with understanding where `Illuminate\Http\Request` is implemented. 

image
image

It looks like `[_ide_helpers.php](https://github.com/inertiajs/inertia-laravel/blob/1.x/_ide_helpers.php)` is propagating its own classes into projects using Inertia. 

I believe that IDE helpers could be great if someone needs them, but they should be in scope of package only. They should not interfere with projects using these packages.

This should fix the problem.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,6 +7,7 @@
 /.gitignore export-ignore
 /.github export-ignore
 /.php-cs-fixer.dist.php export-ignore
+/_ide_helpers.php export-ignore
 /phpunit.xml.dist export-ignore
 /tests export-ignore
 /CHANGELOG.md export-ignore


SamMakesCode:
Yeah, having issues with this too. This feels like a developer preference and shouldn't be included in the repository or it's scope should be limited.

pascalbaljet:
Closing this one in favor of #740 (I'll add you as co-author!)

PR #658, opened 2024-09-06
Update versioning in middleware

RobertBoes:
The first check that's done in the default `version` method of the middleware is to check if `app.asset_url` exists. This was added to support Laravel Vapor, see the comment in the original commit by @claudiodekker: https://github.com/inertiajs/inertia-laravel/commit/b1a4af2094bee755a154ed233f48b3fe825b1e39

This can however introduce unintended behaviour. For example if you have your assets located on a CDN it would just be a static URL. This causes the version hash to not update, so deploying a new version of your app would cause issues as clients would try to fetch non-existing assets and Inertia not forcing a refresh on the client.

A simple fix is instead of just checking if the config value exists we'd check if the app is running within Vapor (see https://github.com/laravel/vapor-core/pull/164)

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -6,6 +6,7 @@ use Closure;
 use Inertia\Support\Header;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Redirect;
+use Inertia\Support\LaravelVapor;
 use Symfony\Component\HttpFoundation\Response;
 
 class Middleware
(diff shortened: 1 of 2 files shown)


nsanden:
+1

Happened to me. Version was never changing due to ASSET_URL being set, so navigation would break each time I would build. Wreaked havoc!

joetannenbaum:
Makes sense, thanks.

PR #664, opened 2024-09-26
Addition of Inertia::response to handle both json and component 

Sarojb31:
Addition of Inertia::reponse  to allow the single route to behave as API that sends props as JSON response if the header is set to 'Accept:application/json' and behave exact to render function if the following header is not matched


joetannenbaum:
Closing this as we're not adding any new features or non-emergency fixes for 1.x. If this is still an issue in 2.x please submit a new PR.

PR #673, opened 2024-10-11
[2.x] Support SSR in Vite dev mode

jrmajor:
Laravel part of inertiajs/inertia#2014.

If user doesn't configure SSR in Vite, the `[vite]/render` endpoint will return 404. In that case, the response is ignored and Inertia proceeds without SSR, so this change should be backwards compatible. It slows the request down a bit, but this happens only in dev (when `laravel-vite-plugin`'s hotfile is present), and only on inital request (consecutive ones aren't SSR-ed anyway) so I think it's acceptable.

I will add tests after the initial idea is approved by maintainers.


pascalbaljet:
Thanks for this! I've refactored it a bit and added tests for `HttpGateway` that were missing anyway. Still need to review the JS counterpart :)

pascalbaljet:
I'm closing this one for the same reason as the [JS counterpart](https://github.com/inertiajs/inertia/pull/2014#issuecomment-3751588323).

PR #677, opened 2024-10-18
[2.x] Fix clearing history when redirecting

joetannenbaum:
This now stores the flag to clear the history in the session so that the _next_ request will clear the history, even if the next request is the result of a redirect.


reinink (approved):
Perfect 👍

PR #678, opened 2024-10-24
[2.x] Ensure closures are resolved via App in resolveArrayableProperties

jaspertey:
This PR resolves a regression in 2.x where shared data closures with arguments could not be resolved. This patch ensures that the closures are resolved using `App::call()`.

e.g.,
```php
Inertia::share('query', fn (Request $request) => $request->query());
```

I encountered this while trying to upgrade a project to Inertia 2.x. A supporting test has been created to demonstrate further.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -173,7 +173,7 @@ class Response implements Responsable
             }
 
             if ($value instanceof Closure) {
-                $value = $value();
+                $value = App::call($value);
             }
 
             if ($unpackDotProps && str_contains($key, '.')) {
--- a/tests/ResponseFactoryTest.php
+++ b/tests/ResponseFactoryTest.php
@@ -146,7 +146,7 @@ class ResponseFactoryTest extends TestCase
         ]);
     }
 
-    public function test_shared_data_can_resolve_closures_arguments(): void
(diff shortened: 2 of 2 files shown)


gainlinejono:
This resolves an issue with the gretel breadcrumbs package, could this be merged? @reinink

jaspertey:
> This resolves an issue with the gretel breadcrumbs package, could this be merged? @reinink

This was my exact use case as well, good to know I wasn't the only one.

wizzymore:
The problem is that - that closure call should not happen here as it is already happenning in the "resolvePropertyInstances" method, and there it works properly. This bug was added in the following commit: https://github.com/inertiajs/inertia-laravel/commit/141256b2ed1833158852c9d239d00abec8ff4942 and the change for the Response.php file should be reverted I think it was a left-over from some tests

jaspertey:
I see that v2.0.0 was released, and the changes in this PR were manually applied in https://github.com/inertiajs/inertia-laravel/commit/75410d7487b3ce9e11d3bddc81f0f7e7c398990e - so I will close this PR as it is officially part of 2.x and no longer relevant!

reinink:
@JasperTey hey, thanks anyway! 🙏

PR #688, opened 2024-11-20
Don't evaluate lazy properties when not specified by partial reload

joetannenbaum:
This PR fixes a situation where properties such as this:

```php
inertia('Users/Index', [
    'users' => fn() => Users::all(),
]);
```

were being evaluated on partial reloads even when `users` was not specified by `X-Inertia-Partial-Data`.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -133,27 +133,41 @@ class Response implements Responsable
      */
     public function resolveProperties(Request $request, array $props): array
     {
-        $isPartial = $this->isPartial($request);
+        $props = $this->resolvePartialProperties($props, $request);
+        $props = $this->resolveArrayableProperties($props, $request);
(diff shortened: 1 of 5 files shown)


joetannenbaum:
Fixes https://github.com/inertiajs/inertia/issues/2094

RobertBoes:
Tested this locally in my demo repo, seems to work as intended, nice! 🚀

jamesparkes:
make it so

hotmeteor:
@joetannenbaum @reinink Would love a merge on this little guy

hschimpf:
I was debugging my app because I was seeing extra hits to the database, then I found this exact behavior.

We really want this _sneaky critter_ merged #up 🚀

_edit: branch tested and works! thanks @joetannenbaum_

PR #692, opened 2024-11-28
[1.x] Ensure Inertia Response generate also compatible with Inertia.js 2

crynobone:
This would allow the application to use multiple packages that use both Inertia.js 1 and 2, relying on `inertia/inertia-laravel` v1 before fully migrating to v2.

## Proposal

With these changes, developers can now update their packages to use Inertia.js v2 without including any new specific v2 features, including Defer props, Polling, etc., and set the packages composer.json to `"inertia/inertia-laravel": "^1.4|^2.0"`. This way, applications that still run on Inertia.js v1 can get the latest package releases without being locked into running older package versions or having to upgrade the application to v2 straight away.

Package developers can choose to ignore this options and enforce `"inertia/inertia-laravel": "^2.0"` to utilise v2 features.

### Pros

* Package developers can safely update npm dependencies to v2.
* Application developers can get updated packages with the latest v2 without changing their code instead locked to older release due to `"inertia/inertia-laravel"` version conflict.

### Cons

* Package developers cannot utilize new Inertia 2 features. 
* Encode histories are ignored until `inertiajs/inertia-laravel` has been updated to v2.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -95,6 +95,7 @@ class Response implements Responsable
             'props' => $props,
             'url' => Str::start(Str::after($request->fullUrl(), $request->getSchemeAndHttpHost()), '/'),
             'version' => $this->version,
+            'encodeHistory' => false,
         ];
 
         if ($request->header(Header::INERTIA)) {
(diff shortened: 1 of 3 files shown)


joetannenbaum:
To clarify this PR after speaking with @crynobone, while the current situation wouldn't actually break anything, if we ever decided to strictly check for a boolean on either of these properties it would break anyone using the `1.x` backend library. So we're just putting up some guardrails for the future.

PR #696, opened 2024-12-17
[2.x] Call `toArray()` on `Arrayable` props resolved from the Container

pascalbaljet:
There's a bug in v2.0 in which `Arrayable` instances that are resolved from the container, for example with deferred props, are not correctly resolved because `toArray()` is never called on them. This PR fixes that and adds a test for it.

In the scenario below, both `Users` and `Departments` implement the `Arrayable` interface, but when resolving `departments` on a partial request, it currently returns an empty object.

```php
return inertia('Users', [
    'users' => Users::make(),
    'departments' => Inertia::defer(fn () => Departments::make()),
]);
```

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -263,6 +263,10 @@ class Response implements Responsable
                 $value = App::call($value);
             }
 
+            if ($value instanceof Arrayable) {
+                $value = $value->toArray();
+            }
+
             if ($value instanceof PromiseInterface) {
                 $value = $value->wait();
             }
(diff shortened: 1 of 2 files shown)


jaspertey:
+1 for this, as I also encountered it in one of my projects during a V2 upgrade.

stubbo:
Any update on this?
Having an issue which will be fixed by changing to v2 but can't upgrade until this is resolved

joetannenbaum:
Looks good, thank you!

Issue #659, opened 2024-09-07
Simpler and fluent syntax to test for props


LemarinelNet:
Hi

I always struggle to remember the correct syntax to test for props in my inertia responses.
Sometimes Copilot helps me a bit, but it also seems to struggle remembering ;)

Since the whole majority of my tests consist of testing if the correct component is returned, but also get props and assert against their value, I wrote two helper functions in my pest testcase : 
```

function getInertiaProps($response)
{
    return json_decode(json_encode($response->viewData('page')), true)['props'];
}

function getInertiaPropKey($response, $key)
{
    return Arr::get(getInertiaProps($response), $key);
}


```

I mostly use the second, like this : 
```
test('I can fetch an order', function () {
    $order = Order::factory()->create();
    OrderLine::factory(5)->create(['order_id' => $order->id]);

    $response = get(route('orders.edit', ['order' => $order->id]));

    expect($response->status())->toBe(200)
        ->and(getInertiaPropKey($response, 'order.lines'))->toBeArray()->toHaveCount(5)
        ->and(getInertiaPropKey($response, 'order.id'))->toBe($order->id);
});

```

It would have be even greater if I did extend the response class so I could write `expect($response->inertiaProp('order.lines))->toBeArray`, but I haven't looked at how to do it.

Am I the only one who finds the `->assertInertia` a bit too complex to write ?
Perhaps there might be another simpler syntax that I missed ?

pascalbaljet:
The upcoming release of the Laravel adapter comes with an `inertiaProps()` method (see PR #700). That makes it much easier to test such data structures with Pest.

Issue #589, opened 2024-03-01
fails testing nested pros


emacaste:
Hi folks, i have some problems with testing. It seems that pest can not handle nested props as documented.
Here my test.

Any ideas?

```php
test('screenings without query sorting are sorted by last screening date desc', function () {
    $screeningDates = [
        'firstDates'  => [
            '2024-01-01 18:00:00',
            '2024-01-02 21:00:00',
            '2024-01-03 21:00:00',
        ],
        'secondDates' => [
            '2024-01-02 18:00:00',
            '2024-01-05 21:00:00',
            '2024-01-06 21:00:00',
        ],
        'thirdDates'  => [
            '2024-01-01 21:00:00',
            '2024-01-02 21:00:00',
            '2024-01-06 21:00:00',
        ]
    ];

    $screenings = [];
    foreach ($screeningDates as $var => $screeningDate) {
        $screenings[$var] = Screening::factory()->create([
            'title' => $var
        ])
            ->each(function (Screening $screening) use ($screeningDate) {
                foreach ($screeningDate as $date) {
                    $screening->dates()->save(ScreeningDate::factory()->make([
                        'date_time' => Carbon::parse($date)
                    ]));
                }
            });
    }

    $this->withoutMiddleware();
    $response = $this->get(route('admin.web.screening.index'));
    $response->assertInertia(fn(AssertableInertia $inertia) => $inertia
        ->component('Admin/Web/Screenings/Index')
        ->has('screenings', 3, fn(AssertableInertia $inertia) => $inertia
//            ->dd('title') // OK
//            ->has('title') // Unexpected properties were found in scope [screenings.0].
//                            Failed asserting that two arrays are identical.
        )
        ->where('screenings.0.title', 'firstDates') // OK
    );
});
```

pascalbaljet:
The upcoming release of the Laravel adapter comes with an `inertiaProps()` method (see PR #700). That makes it much easier to test such data structures with Pest.

PR #700, opened 2024-12-24
[2.x] Introduce inertiaProps Method in TestResponseMacros for Improved Inertia.js Testing

ashokbaruaakas:
This commit introduces an enhanced **inertiaProps** method in the TestResponseMacros class, accompanied by comprehensive tests. The method streamlines the retrieval of specific properties from an Inertia.js response during testing. By leveraging Laravel's `Arr::get` utility, it supports accessing nested properties using dot notation and gracefully handles cases where the specified property is missing by returning a default value.

### Feature
- Direct Prop Extraction: Easily extract all props from the response.
- Dot Notation Support: Access nested prop values using dot notation without causing exceptions.

### Example usage:
`$response->inertiaProps(); // Get All Props`
`$response->inertiaProps('user.name'); // Get Specific Props Value`

This improvement enhances test readability and simplifies property assertions for developers working with Inertia.js.

Contribution of [#659](https://github.com/inertiajs/inertia-laravel/issues/659#issue-2511699537), [#589](https://github.com/inertiajs/inertia-laravel/issues/589#issue-2163491886)

### DIFF
--- a/src/Testing/TestResponseMacros.php
+++ b/src/Testing/TestResponseMacros.php
@@ -3,6 +3,7 @@
 namespace Inertia\Testing;
 
 use Closure;
+use Illuminate\Support\Arr;
 
 class TestResponseMacros
 {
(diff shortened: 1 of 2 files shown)


ashokbaruaakas:
### Example use cases...
image

pascalbaljet:
I think this is a good idea! This would definitely help with testing, but I feel there’s a need for a more extensive way of testing props in general (like testing Lazy props in #595). So I'll keep this one open for a bit while considering a unified approach for this, as I don't want to end up with too many ways of testing props.

pascalbaljet:
Thanks!

PR #704, opened 2025-01-09
Enhance response handling to include objects with toResponse method

antoniopaisfernandes:
Hi,

Would be possible for adding this improvement in Inertia to support props that implement `toResponse` method?

Thank you so much for your help


joetannenbaum:
Thanks for the PR! This is now handled in https://github.com/inertiajs/inertia-laravel/pull/722, closing this one.

PR #705, opened 2025-01-13
Handle SSR URLs with trailing slashes

simon-tma:
If `inertia.ssr.url` ends with a `/`, then generated URL ends with `//render` which results in a 404 response from the SSR server which then causes a very confusing error message in Laravel (access to undefined index "head").

Instead, remove a trailing slash from the URL before appending `/render`.


joetannenbaum:
Looks good, thank you 👍

PR #708, opened 2025-01-26
[1.x] Supports Laravel 12

crynobone:


crynobone:
https://packagist.org/packages/inertiajs/inertia-laravel/php-stats#1

Notes: PHP 7.3 has 0% downloads for a couple of months while PHP 7.4 downloads only max around 0.4%

.github/workflows/tests.yml
```diff
@@ -13,37 +13,50 @@ jobs:
...
         include:
           - php: 8.4
             laravel: 11
             stability: "prefer-stable"
+          - php: 8.4
+            laravel: 12
+            stability: "prefer-stable"
         exclude:
+          - php: 7.3
+            laravel: 8 # Failed security advisory
```
crynobone, line 27: Failed security advisory due to unable to match any fixes https://github.com/advisories/GHSA-c2pc-g5qf-rfrf on PHP 7.3

RobertBoes:
Does Inertia 1.x still need maintenance? The v1 docs are pretty much hidden, it's only linked on the upgrade guide (which is kinda the last place to look at when you're looking for older docs), so I'd assume v1 only receives bug fixes and not updates?

pascalbaljet:
> Does Inertia 1.x still need maintenance? The v1 docs are pretty much hidden, it's only linked on the upgrade guide (which is kinda the last place to look at when you're looking for older docs), so I'd assume v1 only receives bug fixes and not updates?

I recently [asked on X](https://x.com/pascalbaljet/status/1877499727790448791), and many people are still on v1. If it's just a version bump in the composer file, it's not such a big deal, right?

RobertBoes:
> I recently [asked on X](https://x.com/pascalbaljet/status/1877499727790448791), and many people are still on v1. If it's just a version bump in the composer file, it's not such a big deal, right?

A version bump doesn't seem like much on the surface, but that would imply 1.x is actively maintained. If that's the case then I'd also expect fixes to be back-ported to 1.x, which as far as I know isn't the case right now.

I also get that people still use 1.x, so they're not using the current active version. So why should they be pushed to upgrade to Laravel 12, but not to Inertia 2.x? Maintaining multiple versions, for whatever reason, adds to the maintenance. I don't see the point of supporting Laravel 12, as Laravel 11 is still supported to until February 3, 2026. So that's plenty of time to upgrade?

crynobone:
You can use `inertiajs/inertia-laravel` v1 with Inertia.js 2 (npm). This is possible for Laravel packages to get updated to Inertia.js v2 while the application is stucked with v1.

This has no relation to Inertia.js v1 support policy.

pascalbaljet:
> So why should they be pushed to upgrade to Laravel 12, but not to Inertia 2.x?

The goal of Laravel 12 is to be an upgrade with no breaking changes. As much as I love all the work put into Inertia v2, there are still some bugs that prevent people from going into production with it. So, I can imagine people wanting to jump on L12 to get the benefits of new features, but at the same time, they are hesitant to upgrade to Inertia v2.

PR #714, opened 2025-02-11
[2.x] Allow environment config for `ssr.enabled`, `ssr.url`, and `history.encrypt`

bram-pkg:
It is great to be able to configure commonly changed settings through environment variables, as recommended by the 12 factor app guidelines: <https://12factor.net/config>.

In recent times the amount of Laravel config values exposed through environment variables has also increased by a lot.

It only makes sense to also allow for this same practice with Inertia.

### DIFF
--- a/config/inertia.php
+++ b/config/inertia.php
@@ -21,9 +21,9 @@ return [
 
     'ssr' => [
 
-        'enabled' => true,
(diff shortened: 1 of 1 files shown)


joetannenbaum:
Looks good, added `INERTIA_ENCRYPT_HISTORY` in there as well. Thanks!

bram-pkg:
Thanks Joe!

PR #715, opened 2025-02-13
Give render function support for BackedEnum component names

Harrald:
```php
enum PageComponentEnum: string 
{
    case Camera = 'camera';
}


// Make it possible to do this:
return Inertia::render(PageComponentEnum::Camera);

// Instead of this:
return Inertia::render('Camera');
```

### DIFF
--- a/src/ResponseFactory.php
+++ b/src/ResponseFactory.php
@@ -2,6 +2,7 @@
 
 namespace Inertia;
 
+use BackedEnum;
 use Closure;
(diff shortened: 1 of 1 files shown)


joetannenbaum:
I'm not sure this is a common enough use case to merit adding the maintenance overhead, closing for now.

Harrald:
You can find this everywhere in Laravel. So i think it makes sense that the inertia Laravel adapter follows this convention. Right?

Issue #718, opened 2025-02-25
Inertia Testing Fails in GitHub Actions Due to Laravel 12 and Starter Kit Structure


rdehnhardt:
**Description**

With the new Laravel 12 version, Inertia's testing configuration variable `inertia.testing.view-finder` defaults to `resource_path('js/Pages')`. However, the latest starter kits use the `js/pages` directory structure.

This discrepancy causes tests to fail in GitHub Actions, even though they pass locally.

**Steps to Reproduce**

1.  Create a new Laravel 12 project using a starter kit with Inertia and React.
2.  Run tests locally (`./vendor/bin/pest or php artisan test`). (Tests pass)
3.  Commit and push the changes to a GitHub repository with GitHub Actions configured to run tests.
4.  Observe the test failure in GitHub Actions:

    ```
    FAILED  Tests\Feature\Auth\LoginTest > it can render the login screen
    AssertionFailedError
    Inertia page component file [auth/login] does not exist.
    at vendor/inertiajs/inertia-laravel/src/Testing/AssertableInertia.php:63
    ```

**Expected Behavior**

Tests should pass in both local and GitHub Actions environments without requiring manual configuration changes.

**Actual Behavior**

Tests fail in GitHub Actions due to the `inertia.testing.view-finder` configuration mismatch.

**Possible Cause**

The default `inertia.testing.view-finder` configuration in Laravel 12 does not align with the directory structure used by the latest starter kits.

**Additional Information**

* Laravel version: 12
* Inertia version: 2
* Starter Kit: React

**Possible Solutions**

* Update the default `inertia.testing.view-finder` configuration in Laravel 12 to match the starter kit structure.
* Update the starter kits to use the `resource_path('js/Pages')` directory structure.
* Provide clear documentation on how to configure `inertia.testing.view-finder` when using starter kits with Laravel 12.

**Environment details**

* Local Environment: MacOS
* Github Actions: ubuntu-latest

rdehnhardt:
This is the test YML. I've made a few adjustments attempting to resolve the issue, but it largely mirrors the default setup from `laravel new app`.

```yaml
name: tests

on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main

jobs:
  ci:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout
        uses: actions/checkout@v4

      - name: Setup PHP
        uses: shivammathur/setup-php@v2
        with:
          php-version: 8.4
          tools: composer:v2
[...]

PR #719, opened 2025-02-25
fix(config): correct path case in Inertia page_paths

rdehnhardt:
- Updated `page_paths` config to use lowercase 'pages' directory.
- Ensures consistency with naming conventions and avoids errors.

### DIFF
--- a/config/inertia.php
+++ b/config/inertia.php
@@ -47,7 +47,7 @@ return [
 
         'page_paths' => [
 
-            resource_path('js/Pages'),
+            resource_path('js/pages'),
 
         ],
 


RobertBoes:
Personally I don't think this should be changed in the Inertia package. Inertia allows for configuration, so consumers should instead update the paths they use. This would also affect the docs, as that uses the current path. It would also affect other packages, such as Breeze, Jetstream and other third party packages. So changing this default would cause a significantly large breaking change.

rdehnhardt:
I explain a bit better here (https://github.com/inertiajs/inertia-laravel/issues/718) and, in a way, I agree with you. However, as I explain in the issue, the New Laravel Starter Kits have a new standard, and if this is the new standard, every new installation will have to include the config, which I also mention in the issue!

RobertBoes:
Then imho, those starter kits should have the config with the correct path. That's what one would do with a manual install and different paths as well.

That way there would be no breaking change, and it would work in all cases.

With this change it will fix one minor thing, while breaking lots of other implementations.

rdehnhardt:
It's not a problem for me. As I mentioned in the issue, I've already added the configuration to the folder. However, I imagine this problem could be reported many more times in the future! My intention was simply to remove a step of complexity from the inertia installation from now on.

joetannenbaum:
Thank you for the PR, not looking to merge this as users can configure it themselves if it needs to be changed.

PR #722, opened 2025-03-11
fix props that extends Responsable after closures / lazy props

d8vjork:
In a customer app we use a lot the [Laravel Apiable](https://github.com/open-southeners/laravel-apiable) package I was using on Inertia v1 and it worked until we upgraded to Laravel 12 and Inertia v2.

This fixes the problem as Apiable returns a Responsable that was serialised before on Inertia v1 whenever is a lazy prop or not.

It wasn't happening before as it run first the Closure check here:

https://github.com/inertiajs/inertia-laravel/blob/206b1ba0babc1d742308720e766e3b539f39404c/src/Response.php#L153

Then the Arrayable check here converts the result from the Closure to an array:

https://github.com/inertiajs/inertia-laravel/blob/206b1ba0babc1d742308720e766e3b539f39404c/src/Response.php#L173

Now this is separated into 2 blocks breaking this functionality.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -271,6 +271,10 @@ class Response implements Responsable
                 $value = $value->toResponse($request)->getData(true);
             }
 
+            if ($value instanceof Responsable) {
+                $value = $value->toResponse($request)->getData(true);
+            }
+
             if (is_array($value)) {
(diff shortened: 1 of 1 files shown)


joetannenbaum:
Thank you!

PR #732, opened 2025-05-13
[2.x] Allow deepMerge on custom properties

mpociot:
This PR adds the ability to specify custom dot notated keys to use when deep merging data with Inertia.

For example, let's say you have a component that renders a user pagination via infinite loading using the `WhenVisible` component:

```php
Route::get('/users', function () {
        'results' => Inertia::deepMerge(User::paginate()),
    ]);
});
```

This works great while scrolling through the site as new users get appended, however when you perform any additional request that results in a redirect to the `/users` route, the current `deepMerge` strategy will append the users to the state.

This PR aims to fix this by allowing you to specify on which key(s) the data should be merged. In the case of a pagination, you would return:

```php
Route::get('/users', function () {
        'results' => Inertia::deepMerge(User::paginate(), 'data.id'), // <--
    ]);
});
```

This tells Inertia to try and merge the `results.data` array by looking at the `id` property of the underlying objects (if it exists).
If the property is the same, the target object will be replaced with the new one from the response - otherwise it will be appended.

The JS implementation can be found here: https://github.com/inertiajs/inertia/pull/2344

### DIFF
--- a/src/MergeProp.php
+++ b/src/MergeProp.php
@@ -13,11 +13,13 @@ class MergeProp implements Mergeable
 
     /**
      * @param  mixed  $value
+     * @param  string[]  $mergeStrategies
      */
-    public function __construct($value)
+    public function __construct($value, array $mergeStrategies = [])
     {
         $this->value = $value;
         $this->merge = true;
+        $this->mergeStrategies = $mergeStrategies;
     }
 
     public function __invoke()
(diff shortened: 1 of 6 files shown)


pascalbaljet:
Thanks, @mpociot!

pascalbaljet:
Follow-up PR #747

Issue #707, opened 2025-01-19
URL Readability with InertiaJS


muhammedkaraoglu:
### In my project using InertiaJS, I've noticed that square brackets ([]) in URLs are automatically encoded. 
For example:
Original URL:
https://app.test/user?tableFilters[is_active][value]=1
Transformed URL:
https://app.test/user?tableFilters%5Bis_active%5D%5Bvalue%5D=1

While this doesn't cause any functional issues, it does reduce URL readability. Upon investigation, 
I found this is related to the usage of `$request->fullUrl()` in Inertia's `Response`class.

I'd appreciate your thoughts on maintaining more readable URLs. Any insights from developers who have encountered similar situations or have experience with this would be valuable.

```php
/**
     * Create an HTTP response that represents the object.
     *
     * @param  \Illuminate\Http\Request  $request
     * @return \Symfony\Component\HttpFoundation\Response
     */
    public function toResponse($request)
    {
        $props = $this->resolveProperties($request, $this->props);

        $page = array_merge(
            [
                'component' => $this->component,
                'props' => $props,
           **this->     'url' => Str::start(Str::after($request->fullUrl(), $request->getSchemeAndHttpHost()), '/'),
                'version' => $this->version,
                'clearHistory' => $this->clearHistory,
                'encryptHistory' => $this->encryptHistory,
            ],
            $this->resolveMergeProps($request),
            $this->resolveDeferredProps($request),
            $this->resolveCacheDirections($request),
        );

        if ($request->header(Header::INERTIA)) {
            return new JsonResponse($page, 200, [Header::INERTIA => 'true']);
        }

        return ResponseFactory::view($this->rootView, $this->viewData + ['page' => $page]);
    }
```

RobertBoes:
It's a known issue; https://github.com/inertiajs/inertia-laravel/pull/592#issuecomment-1986869324, https://github.com/inertiajs/inertia-laravel/pull/663
Ran into this myself too, but I don't think it'll be fixed in the foreseeable future.

ljbw:
https://github.com/inertiajs/inertia-laravel/pull/731 would fix this issue as well.

pascalbaljet:
As of v2.0.3, you may now use a custom URL resolver that decodes the URL:

```php
Inertia::resolveUrlUsing(function (Request $request) {
    return Str::start(Str::after(rawurldecode($request->fullUrl()), $request->getSchemeAndHttpHost()), '/');
});
```

PR #739, opened 2025-06-05
[2.x] Support for a custom URL resolver

pascalbaljet:
Credit goes to @RobertBoes for bringing [this idea](https://github.com/inertiajs/inertia-laravel/pull/663#issuecomment-2391372314) to the table. I'll add you as an author when this is merged.

With this callback, you can customize the URL that gets sent to the frontend. There are many edge cases and preferences to have one unified way to do this (trailing slashes, encoded query params, etc.). This way, developers can customize the URL to their needs and setup.

```php
Inertia::resolveUrlUsing(function (Request $request) {
    // Return a custom relative URL based on the current request.
    return Str::start(Str::after($request->fullUrl(), $request->getSchemeAndHttpHost()), '/');
});
```

Here are some of the issues/PRs about it:
- #236
- #333
- #359
- #360
- #421
- #422
- #437
- #446
- #499
- #503
- #592
- #663
- #707
- #731

### DIFF
--- a/src/Inertia.php
+++ b/src/Inertia.php
@@ -13,6 +13,7 @@ use Illuminate\Support\Facades\Facade;
  * @method static void flushShared()
  * @method static void version(\Closure|string|null $version)
  * @method static string getVersion()
+ * @method static void resolveUrlUsing(\Closure|null $urlResolver = null)
  * @method static \Inertia\OptionalProp optional(callable $callback)
  * @method static \Inertia\LazyProp lazy(callable $callback)
  * @method static \Inertia\DeferProp defer(callable $callback, string $group = 'default')
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -70,6 +70,16 @@ class Middleware
     }
 
     /**
+     * Defines a callback that returns the relative URL.
+     *
+     * @return Closure|null
+     */
+    public function urlResolver()
+    {
+        return null;
+    }
+
+    /**
      * Handle the incoming request.
      *
      * @return Response
@@ -82,6 +92,10 @@ class Middleware
 
         Inertia::share($this->share($request));
         Inertia::setRootView($this->rootView($request));
+
+        if ($urlResolver = $this->urlResolver()) {
+            Inertia::resolveUrlUsing($urlResolver);
+        }
 
         $response = $next($request);
         $response->headers->set('Vary', Header::INERTIA);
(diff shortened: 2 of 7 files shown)


rojtjo:
In my original PR I also allowed overriding the URL resolver from the middleware. Could we have that added as well?

```php
use Inertia\Middleware;

{
    {
    }
}
```

pascalbaljet:
Yes, I'll work on that! And I'll add you as co-author as well :)

chack1172:
@pascalbaljet there is actually an issue when setting a custom resolver. If the url does not match exactly the url in a <Link>, anchor tags will not work.

For example in the react playground, if you force leading slash:
```
Str::before($request->fullUrl(), '?') . '/'
```

Than go to the home and click article link, it will not redirect to the bottom, unless you change the link adding the trailing slash

pascalbaljet:
> @pascalbaljet there is actually an issue when setting a custom resolver. If the url does not match exactly the url in a , anchor tags will not work.

The custom resolver was not introduced to add query params to URL (or similar changes). Its main goal is to prevent query params from being encoded, like brackets. I'm still playing around with some additional comparisons in the frontend library to improve on this further. Here's an example from the original PR that uses `rawurldecode` on the URL:

```php
Inertia::resolveUrlUsing(function (Request $request) {
    return Str::start(Str::after(rawurldecode($request->fullUrl()), $request->getSchemeAndHttpHost()), '/');
});
```

chack1172:
@pascalbaljet the code I posted was not correct, I wanted to only add a leading slash not adding a query param, I updated it.

mrleblanc101:
@pascalbaljet Why was nothing added to the documentation about this ?

pascalbaljet:
> @pascalbaljet Why was nothing added to the documentation about this ?

It's been released less than an hour ago 😅 I'm working on documenting this and some other new features and doc improvements in general. I'll probably open a PR for that next week.

mrleblanc101:
@pascalbaljet Yeah, I just noticed the doc was a seperate repo.

thaqebon:
It looks like the js core package changes the format of `usePage().url` after the initial page load.
**What I'm trying to do:**
```php
// In HandleInertiaRequests.php
public function urlResolver()
{
    return fn (Request $request) => $request->fullUrl();
}
```

**Expected Behavior:**
I expect `usePage().url` to return the full URL consistently, including the scheme and domain — for example:
`https://app.test/some/page`

**Actual Behavior:**
* On the first page load, `usePage().url` correctly returns:
  `https://app.test/some/page`
* On subsequent client-side navigations, it only returns the path:
  `/some/page`

PR #745, opened 2025-06-15
[2.x] Keep only partial data in mergeProps

vinnygambiny:
This PR resolve this issue https://github.com/inertiajs/inertia/issues/2371

When you reload with partial data, the mergeProps dans deepMergeProps contains all props. So the VueJS will merge the prop with the response, but the response don't have the props.

Example props
```
[
    'user' => $user,
    'foo' => new MergeProp('foo value'),
]
```

If you reload with only `user`, the response will have  `mergeProps: ['foo']` but the `foo` is not there.

The same with the except.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -309,16 +309,16 @@ class Response implements Responsable
 
     public function resolveMergeProps(Request $request): array
     {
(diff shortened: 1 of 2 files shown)


pascalbaljet:
Thanks!

PR #748, opened 2025-06-19
[2.x] Introduce `ProvidesInertiaProps` interface

pascalbaljet:
This PR is an addition to #746 and inspired by [this comment](https://github.com/inertiajs/inertia-laravel/pull/746#issuecomment-2980262905). Given the one-letter difference in interface names, we might have to do some renaming to prevent confusion.

While `Inertia::render()` already support `Arrayable` as props, there's still room left to improve how we organize and reuse props, especially for complex pages. This new `ProvidesInertiaProps` interface has one method that allows you to return *multiple* props from a class, and then use multiple of these classes to gather your props for a page. 

Here's an example of a class implementing this interface:

```php
class LocalizationProps implements ProvidesInertiaProps
{
    public function toInertiaProps(RenderContext $context): iterable
    {
        return [
            'currentLocale' => app()->getLocale(),
            'fallbackLocale' => config('app.fallback_locale', 'en'),
            'availableLocales' => array_keys(config('app.locales', [])),
        ];
    }
}
```

You can use this directly in the `render()` and `with()` methods:

```php
public function index(LocalizationProps $localizationProps)
{
    return Inertia::render('UserProfile', $localizationProps);

    // or...

    return Inertia::render('UserProfile')->with($localizationProps);
}
```

Even cooler, you can combine these prop classes with your other props in an array (without using the spread operator):

```php
Inertia::render('UserProfile', [
    'user' => auth()->user(),
    $localizationProps,
]);
```

For complex pages with many reusable props, this can be a nice cleanup:

```php
public function index(LocalizationProps $localizationProps, PermissionsProps $permissionProps)
{
    return Inertia::render('UserProfile', [
        'user' => auth()->user(),
        $localizationProps,
        $permissionProps,
    ]);

    // or...

    return Inertia::render('UserProfile')
        ->with('user', auth()->user())
        ->with($localizationProps)
        ->with($permissionProps);
}
```

The `RenderContext` object contains the `Request` instance and the component name.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -165,7 +165,7 @@ class Response implements Responsable
                 // Pipe into a Collection to leverage Collection::getArrayableItems()
                 $newProps = array_merge(
                     $newProps,
-                    collect($value->toInertiaProps($renderContext))->all()
+                    collect($value->toInertiaProperties($renderContext))->all()
                 );
             } else {
                 $newProps[$key] = $value;
--- a/src/Response.php
+++ b/src/Response.php
@@ -50,13 +50,15 @@ class Response implements Responsable
     }
 
     /**
-     * @param  string|array  $key
+     * @param  string|array|ProvidesInertiaProps  $key
      * @param  mixed  $value
      * @return $this
      */
     public function with($key, $value = null): self
     {
-        if (is_array($key)) {
+        if ($key instanceof ProvidesInertiaProps) {
+            $this->props[] = $key;
+        } elseif (is_array($key)) {
             $this->props = array_merge($this->props, $key);
         } else {
             $this->props[$key] = $value;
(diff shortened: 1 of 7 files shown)


danmatthews:
Love this @pascalbaljet  - using it with `->with` will be how I use it i'd think, I think it's a nice backwards compatible QOL improvement along with the other PR.

src/Response.php
```diff
@@ -139,13 +142,37 @@ public function resolveProperties(Request $request, array $props): array
...
+        $newProps = [];
+
+        $renderContext = new RenderContext($this->component, $request);
+
+        foreach ($props as $key => $value) {
+            if (is_numeric($key) && $value instanceof ProvidesInertiaProps) {
+                // Pipe into a Collection to leverage Collection::getArrayableItems()
+                $newProps = array_merge(
+                    $newProps,
+                    collect($value->toInertiaProps($renderContext))->all()
```
hosmelq: It would be helpful if you could call this using the container so we can inject dependencies directly into the method.

```php
App::call([$value, 'toInertiaProps'], ['renderContext' => $renderContext])
```

pascalbaljet:
Renamed `ProvidesInertiaProps` to `ProvidesInertiaProperties` + `toInertiaProps` to `toInertiaProperties`

PR #749, opened 2025-06-19
[2.x] Support for checking if the component exists when rendering

### LINKED ISSUE #272: Check if page components exist
Currently the testing framework allows you to test if a page component actually exists. This behaviour is configured in the `/config/inertia.php` file:

```php
'testing' => [
    'ensure_pages_exist' => true,
    'page_paths' => [
        resource_path('js/Pages'),
    ],
    'page_extensions' => ['js', 'svelte', 'ts', 'vue'],
],
```

However, it would be great to do this all the time. Currently if you try to visit a page that has a component that doesn't exist, an exception isn't thrown. All you get is a console error. In production, this is especially problematic, because the user has no idea something went wrong.

Is there any reason why we can't just do this automatically all the time? Yes, there would be an extremely minor performance hit to checking if the file exists. That practically doesn't seem like a real issue.

I'm thinking that we remove the `testing` property from the config, and just have these options available top level. The `ensure_pages_exist` property could accept an array of environments to run this check in:

```php
'ensure_pages_exist' => ['local', 'production', 'testing'], // or set to "true" for always
'page_paths' => [
    resource_path('js/Pages'),
],
'page_extensions' => ['js', 'svelte', 'ts', 'vue'],
```

### LINKED ISSUE #727: View path errors result in impossible to debug frontend errors
How to reproduce:

1. Have any working Inertia rendering route like:

```php
return Inertia::render('Contacts', []);
```

2. Break the view name:
```php
return Inertia::render('ContactZ', []);
```

What I would expect to happen: an error on frontend saying something like `Missing view 'ContactZ'`, or maybe intertia adapter throwing error 500 with the same text.

What actually happens: no server-side errors and this frontend error:

Image

This issue makes refactoring views unnecessarily difficult.

pascalbaljet:
This PR brings support for checking whether a component exists when sending it to the frontend using `Inertia::render()`. This idea was brought up in #272. It uses the same logic as that in the `AssertableInertia` class, but you can configure it separately from each other.

I think I would not enable this in production (though it's always enabled with Blade), but definitely in development. There's no clear indication on the frontend when you pass the wrong component, and I've been bitten by this a few times, so this would be a nice timesaver.

### DIFF
--- a/config/inertia.php
+++ b/config/inertia.php
@@ -31,6 +31,36 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Pages
+    |--------------------------------------------------------------------------
+    |
+    | If you want to ensure that the pages exist, you can set `ensure_pages_exist` to true.
+    | This will throw an exception if the component does not exist on the filesystem
+    | when rendering a page. You may configure this separately for testing.
+    |
+    */
+
+    'ensure_pages_exist' => false,
+
+    'page_paths' => [
+
+        resource_path('js/Pages'),
+
+    ],
+
+    'page_extensions' => [
+
+        'js',
+        'jsx',
+        'svelte',
+        'ts',
+        'tsx',
+        'vue',
+
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
     | Testing
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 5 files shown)


chack1172:
@pascalbaljet I would get the value for config ensure_pages_exist from an env() for who wants to enable it only in development and not in production

pascalbaljet:
> @pascalbaljet I would get the value for config ensure_pages_exist from an env() for who wants to enable it only in development and not in production

Good point, that also nicely aligns with the other config options. Added in 88d55d7

PR #751, opened 2025-06-20
[2.x] Ability to perform a SSR request without a bundle

pascalbaljet:
Fixes issue #694 and builds upon the test from PR #673. 

This PR allows you to enable SSR without the bundle file. This is needed when the SSR server runs on another server or outside of the container. It introduces an `inertia.ssr.dispatch_without_bundle` configuration setting that defaults to `false`. By setting it to `true`, the `BundleDetector` will not be used.

PR #752, opened 2025-06-20
[2.x] Added `inertia:check-ssr` Artisan command and `isHealthy()` method on `HttpGateway`

### LINKED ISSUE #703: [Feature Request] Ability to check if SSR works
Hey, if this is the wrong place to submit feature requests please let me know.

I've finally got Inertia SSR working in Docker and one thing that would be SUPER useful is having a command to check the status/health of SSR. Something such as `php artisan inertia:ssr-ping` or `php artisan inertia:ssr-health`.

I'm not sure how that can be achieved or if its possible nor do I know how to do that but posting here for visibility.

pascalbaljet:
This PR addresses issues #703 and builds upon PR #751 and #673.

The Inertia SSR server has a [`/health`](https://github.com/inertiajs/inertia/blob/c76f35bd4ad9cd3c980cb6812dda51ef9bda5279/packages/core/src/server.ts#L42) endpoint built-in. I've added a `isHealthy()` method to the `HttpGateway` that performs a request against that endpoint. Then there's also an `inertia:check-ssr` Artisan command that's useful for deployments.

### DIFF
--- a/config/inertia.php
+++ b/config/inertia.php
@@ -25,7 +25,7 @@ return [
 
         'url' => env('INERTIA_SSR_URL', 'http://127.0.0.1:13714'),
 
-        'dispatch_without_bundle' => false,
+        'dispatch_without_bundle' => (bool) env('INERTIA_SSR_DISPATCH_WITHOUT_BUNDLE', false),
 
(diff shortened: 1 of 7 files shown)


harrisonratcliffe:
Thank you 🙏

PR #758, opened 2025-07-05
update return type in Middleware share method

hosmelq:
Update the docBlock of the share method in `Middleware` to specify the return type as `array<string, mixed>`, aligning it with the type used in the `middleware.stub` class.


pascalbaljet:
Thanks!

PR #766, opened 2025-07-23
[2.x] Improve inertia helper docblock

mtlukaszczyk:
Improved the docblock for the inertia function, which will now be better processed by tools such as PHPStan. A similar structure is used, for example, in the "response" function in Laravel’s core framework.


pascalbaljet:
Thanks!

PR #770, opened 2025-08-13
[3.x] Add Backed Enum Support to Inertia::render()

onairmarc:
This PR adds backed enum support to `Inertia::render()` through the `ResponseFactory` class.

The render method now first checks if the `$component` parameter is an instance of `BackedEnum`. If `$component` is not a `BackedEnum`, the method continues as normal. If `$component` is a `BackedEnum`, it then checks to ensure that the backing value is a string. If the backing value is a string, the method continues as normal. If the backing value is not a string, the method throws an `InvalidArgumentException` with a message stating the accepted argument types.


onairmarc:
@joetannenbaum can this PR please get some love? 🤓

pascalbaljet:
I like it, but it's a breaking change for people who are extending `ResponseFactory` and overriding the `render()` method. I think the only safe way is to remove the type hint altogether and rely on a docblock.

onairmarc:
@pascalbaljet I just removed the type hint in favor to the docblock

onairmarc:
any movement on this @pascalbaljet?

pascalbaljet:
I'm pushing this to the next major release, just to be sure we're not breaking anyone's app or static analysis because of this change.

PR #771, opened 2025-08-15
[2.x] Add initial props resolver for Initial Page Load

michaelnabil230:
This PR adds support for an `initial` prop option in Inertia.js shared props. It allows developers to define props that should be sent **only with the initial page load** and excluded from subsequent Inertia responses.

This enhancement is particularly useful for values like configuration settings, application metadata, or other static data that doesn't need to be included with every request — reducing payload size and improving response efficiency.

**Example:**

```php
public function initialPropsResolver(): callable
{
    return function (Request $request) {
        return [
            'locale' => 'en',
            'appName' => 'Test App',
            'translations' => [
                'en' => 'English',
                'es' => 'Spanish',
            ],
        ];
    };
}
```

In this example, the defined props (`locale`, `appName`, `translations`) will be shared only on the first Inertia response.

**Why this matters:**

As discussed in [Inertia.js Discussion #993](https://github.com/inertiajs/inertia/discussions/993), there's a recurring need to send certain props — such as app settings, tokens, or translation strings — only with the initial page load, during the app's initial load. This feature provides a clean, built-in solution for that use case.

> [!NOTE]
> If this PR proves useful and is approved for integration, I'd be happy to work on adding the corresponding support in the Inertia.js client-side adapter as well.

See: https://github.com/inertiajs/inertia-laravel/pull/760/

### DIFF
--- a/src/Inertia.php
+++ b/src/Inertia.php
@@ -14,6 +14,7 @@ use Illuminate\Support\Facades\Facade;
  * @method static void version(\Closure|string|null $version)
  * @method static string getVersion()
  * @method static void resolveUrlUsing(\Closure|null $urlResolver = null)
+ * @method static void resolveInitialProps(\Closure|null $initialPropsResolver = null)
  * @method static \Inertia\OptionalProp optional(callable $callback)
  * @method static \Inertia\LazyProp lazy(callable $callback)
  * @method static \Inertia\DeferProp defer(callable $callback, string $group = 'default')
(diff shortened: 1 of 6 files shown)


michaelnabil230:
Hey @pascalbaljet,
Any update on merging this PR?

pascalbaljet:
I'm closing this one in favor of PR #793, which is more extensive and opens up other possibilities as well.

PR #786, opened 2025-11-17
Handle Connection Exceptions During Health Checks

Magnesium38:
The HttpGateway currently determines whether the SSR server is healthy by looking at the status of the response to the health check endpoint. If the server is not available at all though, there will be no response and a ConnectionException will occur. This causes the check-ssr command to fail quite aggressively due to an exception getting thrown without being caught.

Instead of letting that exception fall through, it seems reasonable to rescue the exception in the isHealthy method and say that it's not healthy if an exception occurred. Additionally, I also feel it's reasonable to not report said exception, but I don't have a strong opinion either way. The way that I noticed this was from seeing this exception after deploying while waiting for the SSR server to start up. In that scenario, I'm not concerned with the ConnectionException as the command is the failure indicator that is actually useful.

I had originally wrote the test implementation using `->pushFailedConnection()`, but that was only added in Laravel 11. In order to keep the Laravel 10 support, I took the implementation of the exception from the PR that added it. https://github.com/laravel/framework/pull/53485

### DIFF
--- a/src/Ssr/HttpGateway.php
+++ b/src/Ssr/HttpGateway.php
@@ -61,11 +61,15 @@ class HttpGateway implements Gateway, HasHealthCheck
      */
     public function isHealthy(): bool
     {
-        return rescue(
-            fn () => Http::get($this->getUrl('/health'))->successful(),
-            rescue: false,
-            report: false,
-        );
+        try {
+            return Http::get($this->getUrl('/health'))->successful();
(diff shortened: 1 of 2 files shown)


pascalbaljet:
Thanks! I've refactored it a bit so we can easily swap the static `rejectionForFailedConnection()` method when we drop support for Laravel 10.

PR #788, opened 2025-11-20
Ability to use `<script>` element for initial page data

pascalbaljet:
See: https://github.com/inertiajs/inertia/pull/2687


bram-pkg:
Ah you're way ahead of me! I was going to make this PR after https://github.com/inertiajs/inertia/pull/2687 got merged 😅 thanks!

PR #792, opened 2025-11-24
Handle array inputs in reloadOnly and reloadExcept

hosmelq:
Fix reloadOnly/reloadExcept to accept array inputs.


pascalbaljet:
Thanks!

PR #797, opened 2025-12-12
Support for Flash Data

pascalbaljet:
This PR adds support for Flash Data. This allows you to send data to the frontend that isn't persisted in the browser history state. This makes it ideal for one-time notifications like toasts, highlights, or other transient UI states that shouldn't reappear when navigating back/forward.

You may use `Inertia::flash()` to flash data from the backend.

```php
public function store(StoreUserRequest $request)
{
    $user = User::create($request->validated());

    Inertia::flash('toast', [
        'type' => 'success',
        'message' => 'User created successfully!',
    ]);

    return back();
}
```

You may chain `back()` on `flash()`.

```php
return Inertia::flash('newUserId', $user->id)->back();
```

Or chain `flash()` on `render()` and vice versa.

```php
return Inertia::render('Projects/Index', [
    'projects' => $projects,
])->flash('highlight', $project->id);

// or...

return Inertia::flash()->render();
```

Flash Data is automatically reflashed on redirects, so it reaches the client after the redirect.


mathieutu:
Hey @pascalbaljet I just found about Flash data.

Would you say that we are supposed to totally replace what we usually did using the standard Laravel session, and the Inertia middleware by the usage of `Inertia::flash`, or we should use `Inertia::flash` in the middleware instead of passing the messages as props ?

Screenshot 2026-02-26 at 16 48 07

Screenshot 2026-02-26 at 16 45 47

Thanks!


EDIT For the record:
Just checked the code, and saw how you implemented it. 

I've updated my code to:
Screenshot 2026-02-26 at 17 05 17

and replaced my watcher by a `router.on('flash')`, and it works like a charm.

PR #802, opened 2025-12-27
Fix flash data loss on non-Inertia redirects

TomKita:
When a non-Inertia request sets flash data and redirects to an Inertia page, the flash data was being lost because the middleware returned early before calling `reflash()`.


pascalbaljet:
Thanks!

PR #807, opened 2026-01-14
Fix flash data loss on Inertia::location() (409) responses

sakshamgorey:
Inertia v2.0 uses 409 Conflict for location redirects. Currently, the Middleware only reflashes session data for standard 3xx redirects. This PR ensures that flash data is preserved when using Inertia::location(), maintaining parity with standard redirect behavior.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -122,7 +122,11 @@ class Middleware
         $response = $next($request);
         $response->headers->set('Vary', Header::INERTIA);
 
-        if ($response->isRedirect() || $response->getStatusCode() === 409) {
+        if ($response->isRedirect()) {
+            $this->reflash($request);
+        }
+
+        if ($response->getStatusCode() === 409 && $response->headers->has(Header::LOCATION)) {
             $this->reflash($request);
         }
 
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -122,7 +122,7 @@ class Middleware
         $response = $next($request);
         $response->headers->set('Vary', Header::INERTIA);
 
-        if ($response->isRedirect()) {
+        if ($response->isRedirect() || $response->getStatusCode() === 409) {
             $this->reflash($request);
         }
 
--- a/tests/MiddlewareTest.php
+++ b/tests/MiddlewareTest.php
@@ -13,6 +13,7 @@ use Illuminate\Support\ViewErrorBag;
 use Inertia\AlwaysProp;
 use Inertia\Inertia;
 use Inertia\Middleware;
+use Inertia\SessionKey;
 use Inertia\Tests\Stubs\CustomUrlResolverMiddleware;
 use Inertia\Tests\Stubs\ExampleMiddleware;
 use Inertia\Tests\Stubs\WithAllErrorsMiddleware;
@@ -442,5 +443,25 @@ class MiddlewareTest extends TestCase
                 return Inertia::render('User/Edit', ['user' => ['name' => 'Jonathan']])->toResponse($request);
             });
         });
+    }
+
+    public function test_flash_data_is_preserved_on_inertia_location_responses(): void
+    {
+        Route::middleware([StartSession::class, Middleware::class])
+            ->get('/external-redirect', function () {
+                Inertia::flash('status', 'Redirecting to external site...');
+
+                return Inertia::location('https://google.com');
+            });
+
+        $response = $this->get('/external-redirect', ['X-Inertia' => 'true']);
+
+        $response->assertStatus(409);
+        $response->assertHeader('X-Inertia-Location', 'https://google.com');
+
+        $this->assertEquals(
+            ['status' => 'Redirecting to external site...'],
+            session(SessionKey::FlashData->value)
+        );
     }
 }


sakshamgorey:
I’ve updated the implementation to refine the 409 status check. By verifying the presence of the X-Inertia-Location header alongside the 409 status code, we ensure that the session reflash() is only triggered during legitimate Inertia location redirects

pascalbaljet:
What's the point of using `Inertia::flash()` if you're redirecting out of your app?

sakshamgorey:
Hey @pascalbaljet , Thanks for checking my pr

This isn't about redirecting out of the app. The primary use case is transitioning 
from an Inertia SPA route to a first-party Blade route within the same Laravel 
application (e.g. login, logout, billing, verification).

In Inertia v2, Inertia::location() (409 + X-Inertia-Location) is the official way 
to exit the SPA shell and trigger a full reload. Semantically, this is equivalent 
to a normal Laravel redirect, and flash data is expected to persist.

Currently, flash data is lost because the middleware only reflashes on 3xx responses. 
This PR treats 409 + X-Inertia-Location as a redirect-equivalent and brings 
Inertia::location() to parity with standard Laravel redirect behavior.

pascalbaljet:
I still don't get it. What's the point of having the Inertia Flash Data reflashed to a regular Blade page? Can you give a concrete example of what you're trying to do with the Flash Data on the Blade redirect target?

sakshamgorey:
Hey @pascalbaljet,

## The Use Case

We have a hybrid app where most pages are Inertia, but auth pages and some utility pages are still Blade. Here are two real scenarios where flash data gets lost:

**Logout flow:**
```php
Route::post('/logout', function (Request $request) {
    Auth::logout();
    Inertia::flash('status', 'You have been logged out');
    return Inertia::location('/login'); // Blade page
});
```

```blade
{{-- login.blade.php --}}
@if(session('_inertia_flash_data.status'))
    <div class="alert alert-success">
        {{ session('_inertia_flash_data.status') }}
    </div>
@endif

<form>
    <!-- login form -->
</form>
```

**Report generation:**
```php
Route::post('/reports/generate', function (Request $request) {
    $report = Report::generatePdf($request->filters);
    
    if ($report->size > 10000) {
        Inertia::flash('message', 'Large report generated. Download may take a moment.');
        Inertia::flash('type', 'warning');
    } elseif ($report->hasErrors()) {
        Inertia::flash('message', 'Report generated with some data omitted due to permissions.');
        Inertia::flash('type', 'info');
    } else {
        Inertia::flash('message', 'Your report is ready for download!');
        Inertia::flash('type', 'success');
    }
    
    Inertia::flash('report_id', $report->id);
    
    return Inertia::location('/reports/download'); // Blade page with PDF viewer
});
```

```blade
{{-- reports/download.blade.php --}}
@if(session('_inertia_flash_data.message'))
    <div class="alert alert-{{ session('_inertia_flash_data.type', 'info') }}">
        {{ session('_inertia_flash_data.message') }}
    </div>
@endif

<div class="report-viewer">
    //code here
</div>
```

In both cases, we're using `Inertia::location()` because we need to break out of the SPA and do a full page reload to a Blade template. The Blade template expects to display the flash message with the appropriate context.

Why use Inertia::flash() instead of Laravel's session()->flash()? For consistency - our entire app uses Inertia::flash() for all flash messages across Inertia pages, so continuing to use it when transitioning to Blade pages keeps the API uniform. I know there are other ways to do this (like using Laravel's native flash methods), but sticking with Inertia::flash() means we have one consistent pattern throughout the codebase.

Without this PR, those messages disappear because the middleware only reflashes on 3xx redirects, not 409s.

Let me know if this makes sense or if you need more context!

I can improve the test if you want

sakshamgorey:
Just to clarify and reframe the intent here: this PR is not meant to formalize or encourage Blade templates to depend on `_inertia_flash_data `specifically. Those snippets were only illustrative of the flash state being lost.

The core issue is simpler and framework-level: `Inertia::location()` represents a semantic redirect (it forces a full page reload), but unlike other redirect responses, the middleware currently does not reflash the session. As a result, flash data is dropped prematurely.

Laravel’s invariant is that flash data survives exactly one subsequent request after any redirect-equivalent response. This PR restores that invariant for 409 + X-Inertia-Location without changing any public APIs or expanding flash scope.

How consumers read flash data (Inertia props, Blade, or otherwise) is intentionally left to userland; this change only ensures the session lifecycle behaves consistently with existing redirect handling.

RobertBoes:
Wouldn't this also introduce other implications? For example `Inertia::location()` can also be used to redirect outside of the current app. So if you'd reflash the data, then navigate to something other than your current app and come back to the app, it would show the flash message. Don't think that would be preferable 🤔

pascalbaljet:
I see. In that case, it would be better to use `session()->flash()` as you already know you're moving out of the Inertia app. I think it's even confusing if you're using the internal `inertia.flash_data` key in a Blade part of your app. Looking at the code, I would immediately wonder, why are we using `Inertia::flash()` here while we're redirecting away from Inertia?

Also, I agree with @RobertBoes that this could even cause unwanted side effects.

sakshamgorey:
Thank you for @pascalbaljet  and @RobertBoes taking the time to review this PR and share your feedback. I really appreciate the effort that goes into maintaining this project, and I’m looking forward to contributing to Inertia in the future.

PR #814, opened 2026-01-20
Allow ScrollProp to be optional

gordanielyan:
`Inertia::scroll()` is great for implementing infinite scrolling, but today it always loads as part of the initial response (or immediately after, when deferred).
This makes it awkward to use for **shared data** (e.g. notification list) that:

* should not always be loaded
* should only be fetched when actually needed
* should integrate naturally with `WhenVisible`

This PR adds an `optional()` modifier:

```php
Inertia::scroll(fn () => User::paginate())->optional();
```

### DIFF
--- a/src/LazyProp.php
+++ b/src/LazyProp.php
@@ -5,9 +5,9 @@ namespace Inertia;
 /**
  * @deprecated Use OptionalProp instead for clearer semantics.
  */
-class LazyProp implements IgnoreFirstLoad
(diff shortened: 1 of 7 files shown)


pascalbaljet:
I'm closing this PR as in Inertia 3, the props resolver is rewritten, and you can now use `Inertia::optional(fn () => Inertia::scroll())`. I've added additional tests for the frontend to make sure this works: https://github.com/inertiajs/inertia/pull/3032

Issue #820, opened 2026-01-29
Deferred Properties never resolve in combination with `ProvidesInertiaProperties` class


TimKunze96:
### PHP package version

2.0.19

### Inertia adapter(s) affected (if any)

- [ ] React
- [ ] Vue 3
- [ ] Svelte
- [x] Not Applicable

### Backend stack

Laravel 12
PHP 8.4

### Describe the problem

Inertia::defer() calls defined in `ProvidesInertiaProperties` classes never resolve.

### Steps to reproduce

Create a class that implements `ProvidesInertiaProperties` and defines props with Inertia::defer(). Pass this class to the inertia properties of a controller and try to load them on the frontend inside `<Deferred />` 

https://github.com/TimKunze96/inertia-deferred-bug-provides-properties

PR #821, opened 2026-01-30
Fix inertia prop modifiers being ignored in provider classes

TimKunze96:
This PR addresses the problem outlined in https://github.com/inertiajs/inertia-laravel/issues/820. Deferred props returned from a ProvidesInertiaProperties class were never loaded on the client. The metadata methods (resolveDeferredProps, resolveMergeProps, resolveOnceProps, resolveScrollProps) all read from $this->props to build the frontend metadata, but ProvidesInertiaProperties objects were only expanded into their individual props inside a local variable within resolveProperties() and never written back to $this->props. This meant the frontend never received the deferredProps metadata and never knew to make a follow-up request.

The fix resolves ProvidesInertiaProperties objects into $this->props at the start of toResponse(), before both resolveProperties() and the metadata methods run. The redundant call inside resolveProperties() was removed since it's no longer needed.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -191,6 +191,8 @@ class Response implements Responsable
      */
     public function toResponse($request)
     {
+        $this->props = $this->resolveInertiaPropsProviders($this->props, $request);
+
         $props = $this->resolveProperties($request, $this->props);
 
         $page = array_merge(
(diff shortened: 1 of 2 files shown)


pascalbaljet:
Thanks!

PR #840, opened 2026-02-27
Add a way to manually refresh once props. (#839)

### LINKED ISSUE #839: Once property cannot be refreshed after mutation or on redirect responses
### PHP package version

2.0.21

### Inertia adapter(s) affected (if any)

- [x] React
- [x] Vue 3
- [x] Svelte
- [ ] Not Applicable

### Backend stack

laravel 12.53.0
PHP 8.5

### Describe the problem

When using `Inertia::once()` in shared data (via `HandleInertiaRequests::share()`), two related problems arise that make it impossible to refresh a "once" prop after state changes.

## Context

`Inertia::once()` is useful for sharing data that only needs to be sent once — for example, the authenticated user after login:

```php
// HandleInertiaRequests.php
{
    return [
        ...parent::share($request),
        'auth' => Inertia::once(fn () => ...)),
    ];
}
```

This works perfectly on the initial page load. However, both problems below make it impossible to re-deliver this prop after state changes (e.g. login, logout, profile update).

## Problem : No API to invalidate or re-trigger a "once" prop

There is currently no way to tell Inertia "this `once` prop needs to be sent again on the next response". Once it has been sent, it is gone — even if the underlying data has changed.

This is a significant limitation for shared props that represent mutable state. If a user logs in, logs out, or updates their profile, the `once` prop reflecting that state becomes stale with no built-in mechanism to refresh it.

**There is no equivalent of:**
```php
Inertia::refresh('auth'); // This does not exist yet
```

The only current workaround is to drop `once` entirely and use a regular shared prop, which sends the data on **every** request — defeating the entire purpose of `once`.

## Problem 2: "Once" props are silently lost after a redirect when the destination component differs

When a controller mutates state and returns a redirect, `once` props defined in `share()` are never delivered to the client — and since there is no invalidation API (see Problem 1), there is no way to force them to be re-sent on the next request either.

The `only` mechanism doesn't work in that case, because of the way`\Inertia\Response::isPartial()` is done:

```php
// vendor/inertiajs/inertia-laravel/src/Response.php:762
public function isPartial(Request $request): bool
{
    return $request->header(Header::PARTIAL_COMPONENT) === $this->component;
}
```

When the frontend makes a partial visit (e.g. with `only: ['auth']`) and the server responds with a redirect to a **different component**, `isPartial()` returns `false` on the follow-up request because `X-Inertia-Partial-Component` no longer matches the destination component. The full props are resolved — but `once` props have already been evaluated on the previous request and are not re-evaluated.

**Concrete example:** The `Login` form page makes a partial visit to authenticate, passing `only: ['auth']`. The server logs the user in and redirects to `/` (a `Dashboard` component). On the follow-up request, `isPartial()` returns `false` (correct — it's a different component), but the `user` once-prop is never included the request is not considered `partial`.

```php
// LoginController.php
public function store(Request $request): RedirectResponse
{
    // ... authenticate user ...
    auth()->guard('web')->login($user);
    $request->session()->regenerate();

    return redirect()->intended('/'); // ← user once-prop never reaches the client
}
```

The frontend never receives the updated `auth` prop after login, leaving components that depend on it stale.

## Expected behavior

1. **There should be an explicit API to invalidate/re-trigger a `once` prop**, so that after a state mutation a controller can signal that the prop must be re-sent on the next response:
```php
Inertia::refresh('auth');
```

2. **`once` props should survive redirect boundaries** — if a `once` prop could not be delivered because the response was a redirect (or because the destination component differed from the requesting component), it should be included in the response to the redirect's destination.

Thanks!

### Steps to reproduce

1. Define a `once` shared prop in `HandleInertiaRequests::share()`:

```php
'auth' => Inertia::once(fn () => str_rand()),
```

2. On a login page, make a **partial visit** using `only: ['auth']` (e.g. via a `<Link>` component with `:only="['auth']"`).

3. The server authenticates the user and returns a **redirect** to a different page (e.g. `Dashboard`).

4. Inertia follows the redirect and makes a **full, non-partial request** to the destination page.

**Expected:** `auth` is included in the response since it's a full request and the prop hasn't been delivered yet.

**Actual:** `auth` is absent — the `once` closure was already evaluated and is not re-evaluated on the follow-up request.

---

1. Define the same `once` shared prop.
2. Log the user in via any standard (non-partial) request.
3. Try to force `auth` to be re-sent on the next response from a controller — there is no API to do so.
4. The only workaround is removing `once`, causing the prop to be sent on **every** request.

mathieutu:
Hi,
This PR fixes #839.

It adds a `Inertia::refresh('prop')` method, to force refresh once prop.
Everything is summarized in the last test (of course, consider `sharedData` instead of manual render):

```php
public function test_complete_refresh_example(): void
{
    Route::middleware([StartSession::class, ExampleMiddleware::class])->post('/dashboard', function () {
        return Inertia::render('Dashboard', [
            'foo' => Inertia::once(fn () => 'bar'),
        ]);
    });

    Route::middleware([StartSession::class, ExampleMiddleware::class])->post('/logout', function () {
        Inertia::refresh('foo');

        return redirect('/login');
    });

    Route::middleware([StartSession::class, ExampleMiddleware::class])->get('/login', function () {
        return Inertia::render('Auth/Login', [
            'foo' => Inertia::once(fn () => 'baz'),
        ]);
    });

    $dashboardResponse = $this->post('/dashboard', [], ['X-Inertia' => 'true']);
    $dashboardResponse->assertSuccessful();
    $dashboardResponse->assertJson(['props' => ['foo' => 'bar']]);
    $dashboardResponse->assertSessionMissing(SessionKey::Refresh->value);

    $logoutResponse = $this->post('/logout', [], ['X-Inertia' => 'true']);
    $logoutResponse->assertRedirect('/login');
    $logoutResponse->assertSessionHas(SessionKey::Refresh->value, ['foo']);

    $loginResponse = $this->get('/login', [
        'X-Inertia' => 'true',
        'X-Inertia-Except-Once-Props' => 'foo',
    ]);

    $loginResponse->assertSuccessful();
    $loginResponse->assertJson(['props' => ['foo' => 'baz']]);
    $loginResponse->assertSessionMissing(SessionKey::Refresh->value);
}
```

Thanks for your awesome work!

### DIFF
--- a/src/Inertia.php
+++ b/src/Inertia.php
@@ -27,6 +27,8 @@ use Illuminate\Support\Facades\Facade;
  * @method static \Symfony\Component\HttpFoundation\Response location(string|\Symfony\Component\HttpFoundation\RedirectResponse $url)
  * @method static \Inertia\ResponseFactory flash(string|array<string, mixed> $key, mixed $value = null)
  * @method static array<string, mixed> getFlashed(?\Illuminate\Http\Request $request = null)
+ * @method static \Inertia\ResponseFactory refresh(string|array<string> ...$keys)
+ * @method static array<int, string> getRefreshed(?\Illuminate\Http\Request $request = null)
  * @method static void macro(string $name, object|callable $macro)
  * @method static void mixin(object $mixin, bool $replace = true)
  * @method static bool hasMacro(string $name)
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -153,6 +153,10 @@ class Middleware
         if ($flashed = Inertia::getFlashed($request)) {
             $request->session()->flash(SessionKey::FlashData->value, $flashed);
         }
+
+        if ($refreshed = Inertia::getRefreshed($request)) {
+            $request->session()->flash(SessionKey::Refresh->value, $refreshed);
+        }
     }
 
     /**
--- a/src/Response.php
+++ b/src/Response.php
@@ -304,18 +304,20 @@ class Response implements Responsable
      */
     public function resolveOnceProperties(array $props, Request $request): array
     {
-        if (! $this->isInertia($request) || $this->isPartial($request)) {
+        if (! $this->isInertia($request)) {
             return $props;
         }
 
+        $isPartial = $this->isPartial($request);
         $exceptOnceProps = $this->getExceptOnceProps($request);
+        $refreshedProps = Inertia::getRefreshed($request);
 
         if (count($exceptOnceProps) === 0) {
             return $props;
         }
 
         return collect($props)
-            ->reject(function ($prop, string $key) use ($exceptOnceProps) {
+            ->reject(function ($prop, string $key) use ($isPartial, $exceptOnceProps, $refreshedProps) {
                 if (! $prop instanceof Onceable) {
                     return false;
                 }
@@ -328,7 +330,17 @@ class Response implements Responsable
                     return false;
                 }
 
-                return in_array($prop->getKey() ?? $key, $exceptOnceProps);
+                $resolvedKey = $prop->getKey() ?? $key;
+
+                if (in_array($resolvedKey, $refreshedProps)) {
+                    return false;
+                }
+
+                if ($isPartial) {
+                    return false;
+                }
+
+                return in_array($resolvedKey, $exceptOnceProps);
             })
             ->all();
     }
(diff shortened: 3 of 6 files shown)


pascalbaljet:
I'm closing this PR as I find `Inertia::refresh()` confusing as it's not explicitly for _once props_. Also, let's continue the discussion in the [issue](https://github.com/inertiajs/inertia-laravel/issues/839) first to see if we really need a built-in solution for this. It feels like this belongs more in app code than in the package.

PR #841, opened 2026-03-06
[3.x] Add pages transform callback

nhedger:
### What

This PR adds support for an optional `pages.transform` configuration when resolving component paths during file existence checks, both at runtime and during tests.

### How

This PR allows defining `pages.transform` as a callable that receives the `$name` of the component being resolved and returns the transformed name.

### Why

On the frontend, I configure `pages.transform` to resolve page components so that a name like `home` maps to `home/page`:

```ts
createInertiaApp({
  pages: {
    path: "./pages",
    extension: ".tsx",
    transform: (name) => `${name}/page`,
  },
});
```

However, this transformation cannot currently be expressed on the backend when verifying that a component file exists. As a result, the backend attempts to check for `home.tsx`, while the actual page file is `home/page.tsx`.


nhedger:
@pascalbaljet Could you let let me know if there's a blocker ?

pascalbaljet:
Thanks for the PR! The use case makes sense, but a couple of things:

1. We shouldn't put callables in config files. It should follow the same pattern as `resolveUrlUsing` (see `ResponseFactory`).

2. Rather than only applying the transform during the file existence check, I think we should apply it in `render()` itself, so the resolved name is also what gets sent to the frontend. That way you wouldn't need the frontend `transform` option at all:

```php
Inertia::resolveComponentUsing(fn ($name) => "$name/Page");
```

Would you be up for refactoring the PR to use this approach?

nhedger:
Makes sense, I've refactored the PR to follow your suggested approach.

pascalbaljet:
Thanks! I've renamed the method to `transformComponentUsing()`.

PR #842, opened 2026-03-10
[3.x] fix: use rtrim for SSR shutdown URL construction

Anthony14FR:
## Problem

`StopSsr` builds the shutdown URL using `str_replace('/render', '', ...)`, which is a leftover from when `inertia.ssr.url` pointed directly to the render endpoint. Since v3, `inertia.ssr.url` is a **base URL** with no path.

As a result, a trailing slash in `INERTIA_SSR_URL` produces a double slash in the shutdown URL:

```env
INERTIA_SSR_URL=http://127.0.0.1:13714/
```

```
http://127.0.0.1:13714//shutdown  ❌
```

## Fix

Replace `str_replace('/render', '', ...)` with `rtrim(..., '/')`, which strips any trailing slash before appending `/shutdown`:

```
http://127.0.0.1:13714/   →  http://127.0.0.1:13714/shutdown  ✅
http://127.0.0.1:13714    →  http://127.0.0.1:13714/shutdown  ✅
```

This is consistent with how `HttpGateway::getProductionUrl()` already handles the same config value:

```php
$baseUrl = rtrim(config('inertia.ssr.url', 'http://127.0.0.1:13714'), '/');
```

### DIFF
--- a/src/Commands/StopSsr.php
+++ b/src/Commands/StopSsr.php
@@ -27,7 +27,7 @@ class StopSsr extends Command
      */
     public function handle(): int
     {
-        $url = str_replace('/render', '', config('inertia.ssr.url', 'http://127.0.0.1:13714')).'/shutdown';
+        $url = rtrim(config('inertia.ssr.url', 'http://127.0.0.1:13714'), '/').'/shutdown';
 
         $ch = curl_init($url);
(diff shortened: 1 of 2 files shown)


pascalbaljet:
Thanks!

PR #844, opened 2026-03-13
[3.x] Add @inertiaHeadFallback directive to prevent duplicate title tags on SSR

### LINKED ISSUE #666: Duplicate title tags on SSR
I am using breeze starter kit with Vue and it works great but when I try SSR mode I see 2 title tags in the head.

One tag is from the root view (app.blade.php) and the other is set via the <Head> component.

image

Is there any way to fix this?

hamedelasma:
Fixes #666

## Problem

When using SSR, the `@inertiaHead` directive outputs the server-rendered `<head>` content (which includes a `<title>` tag from the `<Head>` component), but the Blade root view (`app.blade.php`) also contains its own `<title>` tag as a fallback — resulting in duplicate `<title>` tags in the HTML output.

## Solution

Adds a new `@inertiaHeadFallback` / `@endInertiaHeadFallback` Blade directive pair that conditionally renders its content **only when SSR is not active** (i.e., SSR is disabled or the SSR server is down). This follows the approach suggested by @pascalbaljet in #666.

### Usage

```blade
<head>
    @inertiaHead

    @inertiaHeadFallback
        <title>{{ config('app.name') }}</title>
    @endInertiaHeadFallback
</head>
```

- When SSR is **active**: only the SSR-rendered `<title>` appears (no duplicates)
- When SSR is **disabled or unavailable**: the fallback `<title>` renders normally

### Implementation details

- Reuses the existing `$__inertiaSsrDispatched` / `$__inertiaSsrResponse` variables to avoid dispatching to the SSR server more than once per request
- Works with or without `@inertiaHead` preceding it
- 5 new tests covering all scenarios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,17 @@
 # Release Notes
 
-## [Unreleased](https://github.com/inertiajs/inertia-laravel/compare/v2.0.21...2.x)
+## [Unreleased](https://github.com/inertiajs/inertia-laravel/compare/v2.0.22...2.x)
 
 - Nothing!
+
+## [v2.0.22](https://github.com/inertiajs/inertia-laravel/compare/v2.0.21...v2.0.22) - 2026-03-11
+
+### What's Changed
+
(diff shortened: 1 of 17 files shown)


pascalbaljet:
Thanks for this. I'm closing it in favor of #848.

PR #845, opened 2026-03-21
[3.x] Default config page_paths to lowercase

jlswanson28694:
This commit changes the Inertia config file to match the Laravel Inertia starter kit default page components directory of `resources/js/pages`. The Inertia config had `page_paths` and `testing.page_paths` set to include `resource_path('js/Pages')`. This is case-sensitive, resulting in `ensure_pages_exist => true` being unable to find any page components.


pascalbaljet:
Thanks! I'm closing this one in favor of #847 so we can target Inertia 3.0.

PR #852, opened 2026-03-26
Add runtime_path config option for SSR

shaffe-fr:
When using a Node version manager like nvm, the node binary isn't always just `node`. This adds a `runtime_path` config option (`inertia.ssr.runtime_path`) so you can point to a specific node binary when starting the SSR server.

Usage: set `INERTIA_SSR_RUNTIME_PATH` in your `.env` or uncomment the option in `config/inertia.php`.

This could be applied to both 2.x and 3.x.


pascalbaljet:
Any reason this is only for node? (`$runtime === 'node'`)

shaffe-fr:
You're right, I renamed everything to runtime_path.

pascalbaljet:
I'm closing this one in favor of #857, targeting Inertia 3.

shaffe-fr:
👌

PR #854, opened 2026-04-03
[3.x] Fix middleware priority

jlswanson28694:
Fix for PR #801
### Problem
I was pulling my hair out over this and finally figured out what was happening. The `afterResolving` callback in `ServiceProvider::configureMiddlewarePriority` doesn't actually run during a real request. The service provider registers that callback to run `addToMiddlewarePriorityAfter` when `Illuminate\Foundation\Http\Kernel` is resolved from the container, but the Kernel is resolved in `Illuminate\Foundation\Application::handleRequest`, before the application is bootstrapped. The tests are misleading because the application is bootstrapped before any of the tests run.

### Proposed solution
Don't rely on an `afterResolving` callback to set the middleware priority. Just set it right away when the service provider boots.

### DIFF
--- a/src/ServiceProvider.php
+++ b/src/ServiceProvider.php
@@ -72,12 +72,11 @@ class ServiceProvider extends BaseServiceProvider
      */
     protected function configureMiddlewarePriority(): void
     {
-        $kernel = $this->app->make(HttpKernelContract::class);
-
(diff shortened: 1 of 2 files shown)


pascalbaljet:
Thanks!

PR #885, opened 2026-07-14
[3.x] Custom SSR Hot URL

goreks-dev:
## 1. The Problem

When running Laravel inside a Docker container (e.g., Laravel Sail) and Vite on the host machine (or in a separate container), the `Vite::hotFile()` contains the URL resolved from the browser's perspective (e.g., `http://localhost:5173` or `https://vite.app.test`).

```js
// vite.config.js

server: {
    hmr: {
        host: "vite.app.test",
        protocol: "wss",
        clientPort: 443,
    },
    watch: {
        ignored: ["**/storage/**", "**/*.md"],
    },
},

```

When Inertia SSR is running in development mode (hot), the `HttpGateway` attempts to dispatch SSR requests using the URL retrieved from `Vite::hotFile()`:

```php
protected function getHotUrl(string $path = '/'): string
{
    return rtrim(file_get_contents(Vite::hotFile())).$path;
}
```

This causes connection failures inside containerized environments:
- **Loopback Issues:** `localhost` inside the Docker PHP container resolves to the container's loopback interface instead of the host machine (where Vite dev server is running).
- **DNS/SSL Errors:** Custom local domains (e.g., `https://vite.app.test`) may not resolve inside the container, or curl requests will fail due to untrusted local SSL certificates.

Since there is currently no way to configure the Vite dev server URL specifically for SSR requests in `HttpGateway`, developers must override the class manually.

---

## 2. Proposed Solution

Introduce a new hot_url` configuration option under `ssr` in `config/inertia.php` (e.g., pointing to `http://127.0.0.1:5173` or `http://host.docker.internal:5173`). If `hot_url` is provided, `HttpGateway` will use it to route local SSR requests instead of reading `Vite::hotFile()`.

### DIFF
--- a/config/inertia.php
+++ b/config/inertia.php
@@ -29,7 +29,7 @@ return [
 
         'url' => env('INERTIA_SSR_URL', 'http://127.0.0.1:13714'),
 
-        'hot_url' => env('INERTIA_SSR_DEV_URL'),
+        'hot_url' => env('INERTIA_SSR_HOT_URL'),
 
         'ensure_bundle_exists' => (bool) env('INERTIA_SSR_ENSURE_BUNDLE_EXISTS', true),
 
--- a/config/inertia.php
+++ b/config/inertia.php
@@ -29,7 +29,7 @@ return [
 
         'url' => env('INERTIA_SSR_URL', 'http://127.0.0.1:13714'),
 
-        'dev_url' => env('INERTIA_SSR_DEV_URL'),
+        'hot_url' => env('INERTIA_SSR_DEV_URL'),
 
         'ensure_bundle_exists' => (bool) env('INERTIA_SSR_ENSURE_BUNDLE_EXISTS', true),
 
--- a/config/inertia.php
+++ b/config/inertia.php
@@ -29,6 +29,8 @@ return [
 
         'url' => env('INERTIA_SSR_URL', 'http://127.0.0.1:13714'),
 
+        'dev_url' => env('INERTIA_SSR_DEV_URL'),
+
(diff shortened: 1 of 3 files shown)


RobertBoes:
> When running Laravel inside a Docker container (e.g., Laravel Sail) and Vite on the host machine

That does defeat the purpose of using Docker, you have an isolated development environment, so then running commands outside of the container is a recipe to cause issues. This is also noted in the Sail docs, you should run commands inside the container; https://laravel.com/docs/13.x/sail#executing-node-npm-commands, so you'd use `sail npm run dev`

> or in a separate container

I'd say that would be an issue, it's quite typical to have a compose file, running node, php / ngingx in separate containers, so that's worth fixing imho

goreks-dev:
I think there’s a slight misunderstanding—I am running all my commands inside the container!

What I meant is that I use Portless on my host machine as a reverse proxy to map clean domain aliases to the container's exposed ports, like this:

[https://app.localhost](https://app.localhost) -> localhost:9008

[https://vite.app.localhost](https://vite.app.localhost) -> localhost:5173

[https://ws.app.localhost](https://ws.app.localhost) -> localhost:8080

So the environment remains completely isolated inside Docker; Portless just handles the local SSL and domain routing for me.

svmrw:
Apologies, I created a similar [issue](https://github.com/inertiajs/inertia-laravel/issues/887) yesterday. I only just saw this thread.

I created a patch for myself; personally, I’m perfectly happy with this simple solution, and it shouldn't cause any issues for anyone else.

P.S. My own application runs perfectly outside of Sail. I use separate containers for PHP, Node, and MariaDB.
It might be a slightly unconventional approach, but everything works very well.

An intractable problem arose only in development mode with SSR enabled.

pascalbaljet:
Thanks!

PR #888, opened 2026-07-21
Store previous locations for Inertia visits

### LINKED ISSUE #886: previousUri and previousRoute not set on Inertia route visit
> Cross-posted from [inertiajs/inertia#3199](https://github.com/inertiajs/inertia/issues/3199)

## Backend stack

Laravel 13.x
PHP 8.5

## Describe the problem

`previousUri` and `previousRoute` are not set on Inertia route visits (where the page is updated on the client side). The latest page loaded by the browser that did not use Inertia routing is returned as the previous route.

**Expected behavior:** `previousUri` and `previousRoute` should be set on Inertia route visits, just like in traditional routing (Blade).

## Steps to reproduce

1. Create the following routes with their respective controllers and models:

   - `organizations.overview`
   - `organizations.edit`
   - `organizations.users.index`
   - `organizations.users.create`

2. Load/open any of the routes in the browser.
3. Navigate to any other route(s) using Inertia routing (`<Link>`, `router`, etc.).
4. Finally, POST to `organizations.set-active`. The dumped route and URI will be the route from step 1, which is the one loaded without Inertia routing.

```php
Route::post('organizations/{organization}/set-active', function (Request $request, Organization $organization): RedirectResponse {
    $previousRoute = $request->session()->previousRoute();
    $previousUri = $request->session()->previousUri();

    dd($previousRoute, $previousUri);

    // Below is a little example of what can be done with this feature

    // $request->session()->put('organization_id', $organization->id);

    // if (is_null($previousRoute) || $previousRoute === 'organizations.overview') {
    //     return to_route('organizations.overview', ['organization' => $organization]);
    // }

    // if (in_array($previousRoute, ['organizations.edit', 'organizations.users.index', 'organizations.users.create'])) {
    //     return to_route($previousRoute, ['organization' => $organization]);
    // }

    // return to_route('organizations.overview', ['organization' => $organization]);
})->name('organizations.set-active');
```

drewmt:
## Summary

Inertia page visits include the `X-Requested-With: XMLHttpRequest` header, so Laravel's session middleware intentionally skips updating the previous URL and route. As a result, session navigation history can remain stuck on the last full-page visit.

This change stores the current URL and route for Inertia GET visits that Laravel skips, while preserving the existing exclusions for regular AJAX requests, prefetches, precognitive requests, and non-GET requests. Route history is detected as an optional session capability so Laravel 11 remains supported.

Fixes #886.

## User impact

Applications can reliably use `previousUrl()` / `previousUri()` and, on supported Laravel versions, `previousRoute()` after client-side Inertia navigation. Existing non-navigation requests keep their current behavior.

## Validation

- Added five middleware regression tests covering a real Inertia visit and the AJAX, prefetch, precognitive, and non-GET boundaries.
- Confirmed the main regression test fails on the current `3.x` implementation before the fix.
- Laravel 11.55: 413 tests, 1,513 assertions.
- Laravel 12.64: 413 tests, 1,513 assertions; full PHPStan passes.
- Laravel 13.21: 413 tests, 1,513 assertions.
- Pint and PHPStan pass for all changed files across the tested framework versions.
- `composer validate --strict` and `composer audit` pass with current dependencies.

### DIFF
--- a/src/Middleware.php
+++ b/src/Middleware.php
@@ -3,7 +3,9 @@
 namespace Inertia;
 
 use Closure;
+use Illuminate\Contracts\Session\Session as SessionContract;
 use Illuminate\Http\Request;
+use Illuminate\Routing\Route;
 use Illuminate\Session\Store;
 use Illuminate\Support\Facades\Redirect;
 use Illuminate\Support\MessageBag;
@@ -146,6 +148,8 @@ class Middleware
             return $response;
         }
 
+        $this->storeCurrentUrl($request);
+
         if ($request->method() === 'GET' && $request->header(Header::VERSION, '') !== Inertia::getVersion()) {
             $response = $this->onVersionChange($request, $response);
         }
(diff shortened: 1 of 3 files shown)


pascalbaljet:
Thanks for the PR! I've pushed some changes to your branch:

* The behavior is now behind a new `store_previous_url` config option, as it changes what `previousUrl()` returns in existing apps.
* Partial reloads are excluded, as those aren't navigations. You can override `shouldStoreCurrentUrl()` in your middleware to change this.

PR #894, opened 2026-07-29
[3.x] Fix DevTools recording under Laravel Octane

### LINKED ISSUE #893: Inertia.js DevTools Doesn't Work With Laravel Octane
### PHP package version

3.2.0

### Inertia adapter(s) affected (if any)

- [ ] React
- [ ] Vue 3
- [ ] Svelte
- [x] Not Applicable

### Backend stack

- PHP 8.5+
- Laravel 13+
- PostgreSQL 17+
- Inertia.js 3.0+

### Describe the problem

Inertia.js DevTools recorder does not work when you are running the project with Laravel Octane because of the `terminating` callback used in https://github.com/inertiajs/inertia-laravel/blob/3.x/src/DevTools/DevToolsServiceProvider.php#L33-L43 to flush the events. I had to manually register a `RequestHandled` listener that runs that exact code before it could work:

```php
<?php

declare(strict_types=1);

namespace App\Listeners;

use Inertia\DevTools\DevTools;
use Inertia\DevTools\EntryStore;
use Inertia\DevTools\EntriesRepository;

class FlushInertiaDevtoolsStore
{
    /**
     * Handle the event.
     */
    public function handle($event): void
    {
        if (! DevTools::enabled()) {
            return;
        }

        $repository = $event->sandbox->make(EntriesRepository::class);

        $event->sandbox->make(EntryStore::class)->flush($repository);

        $repository->pruneIfDue();
    }
}
```

### Steps to reproduce

- Set up Laravel Octane with any of the drivers (I am using Frankenphp)
- Set up the Inertia.js DevTools extension
- Start the Octane server
- Open the project in the browser
- You should not see any events

lazerg:
The DevTools recorder flushes its collected entries from an `app->terminating()` callback registered in the service provider. Under Octane the application instance is reused across requests and that callback never fires per request, so nothing is written to storage and the extension shows no entries. Moving the flush to a `RequestHandled` listener, resolved from the active container, persists the entry at the end of every request under both Octane and a normal FPM setup.

Fixes #893

### DIFF
--- a/src/DevTools/DevToolsServiceProvider.php
+++ b/src/DevTools/DevToolsServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace Inertia\DevTools;
 
+use Illuminate\Foundation\Http\Events\RequestHandled;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\ServiceProvider;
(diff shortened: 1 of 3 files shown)


pascalbaljet:
Thanks!

devhammed:
@lazerg Thanks for this, I didn't know there is a built-in RequestHandled event in Laravel!

🙏🙏🙏

PR #907, opened 2026-09-02
Fix loadDeferredProps() crashing when a group name matches a helper function

mdalikadar:
`loadDeferredProps()` uses `is_callable()` to check if you passed a closure or a group name. The problem is `is_callable('auth')` returns `true`, because `is_callable()` on a string just checks if a function with that name exists — and Laravel has a global `auth()` function.

So if you name a deferred prop group something like `'auth'`, `'session'`, or `'collect'` (all totally normal group names), this:

```php
$page->loadDeferredProps('auth', function ($page) {
    $page->where('user', $user);
});
```

crashes with:

```
TypeError: reloadOnly(): Argument #2 ($callback) must be of type ?Closure, string given
```

because it silently treats `'auth'` as the callback and drops your real closure.

The fix just checks `instanceof Closure` instead, which is what the method's own type hint (`Closure|array|string`) already says it should be.

Added a test that reproduces the crash with a group named `auth` and confirms it's fixed. Full test suite, Pint, and Larastan all pass.

### DIFF
--- a/src/Testing/AssertableInertia.php
+++ b/src/Testing/AssertableInertia.php
@@ -141,9 +141,11 @@ class AssertableInertia extends AssertableJson
      */
     public function loadDeferredProps(Closure|array|string $groupsOrCallback, ?Closure $callback = null): self
     {
-        $callback = is_callable($groupsOrCallback) ? $groupsOrCallback : $callback;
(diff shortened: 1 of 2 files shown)


mdalikadar:
Closing to reopen as a fresh PR without edit history.

PR #910, opened 2026-09-04
[3.x] Return Illuminate's RedirectResponse from back()

YannikFirre:
`ResponseFactory::back()` is declared as returning `Symfony\Component\HttpFoundation\RedirectResponse`, but it just returns `Redirect::back(...)`, which is an `Illuminate\Http\RedirectResponse`. The facade docblock in `Inertia.php` repeats the Symfony type.

In practice this bites any controller that's typed the Laravel way:

```php
public function update(Request $request): \Illuminate\Http\RedirectResponse
{
    // ...
    return Inertia::flash('toast', ['message' => __('Saved.')])->back();
}
```

PHPStan (level 8 here, but any level that checks return types) reports `Method ... should return Illuminate\Http\RedirectResponse but returns Symfony\Component\HttpFoundation\RedirectResponse`, so the chained form ends up being avoided in favour of `redirect()->back()` plus a separate `Inertia::flash()` call. Runtime was always fine; only the declared type was too wide.

This narrows the return type to the class that's actually returned (the Illuminate one extends the Symfony one, so nothing that accepted the old type breaks) and fixes the `@method` line on the facade to match. `location()` is untouched: it genuinely accepts and returns the Symfony types.

No behaviour change. `vendor/bin/phpunit` passes (560 tests), `pint` is clean, and `phpstan` reports the same single pre-existing warning in `ExceptionResponse.php:157` before and after this change.

### DIFF
--- a/src/Inertia.php
+++ b/src/Inertia.php
@@ -30,7 +30,7 @@ use Illuminate\Support\Facades\Facade;
  * @method static \Symfony\Component\HttpFoundation\Response location(string|\Symfony\Component\HttpFoundation\RedirectResponse $url)
  * @method static void handleExceptionsUsing(callable $callback)
  * @method static \Inertia\ResponseFactory flash(\BackedEnum|\UnitEnum|string|array<string, mixed> $key, mixed $value = null)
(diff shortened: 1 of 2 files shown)


pascalbaljet:
Thanks!

